PR #111, opened 2022-12-01
APP-12576 bridge base

Sqrrl:
## Summary

- [Ticket](https://flipapp.myjetbrains.com/youtrack/issue/APP-12576/Create-basis-for-JS-Bridge)
- see [README](https://github.com/getflip/swirl/blob/APP-12576-bridge-base/packages/bridge/README.md) for details
- new @getflip/bridge package added
- targeted at external integrations
- implements postMessage-based communication between the external integration and one of our host apps
- [JSON-RPC](https://de.wikipedia.org/wiki/JSON-RPC)-inspired message format
- implements a few exemplary functions


webspaceadam (approved):
Played a little bit with it, and very promising. 
Good Job!

image

Sqrrl:
> Played a little bit with it, and very promising. Good Job!
> 
> image

How did you get this error?

webspaceadam:
i implemented it to play with the bridge and get a better feeling for its logic 😅

packages/bridge/src/messaging/messaging.ts
```diff
@@ -0,0 +1,63 @@
+import { log } from "../logging";
+import {
+  BridgeError,
+  BridgeErrorCode,
+  BridgeRequest,
+  BridgeResponse,
+} from "../types";
+
+export function postMessage(message: BridgeRequest) {
```
konnic, line 9: **question:** Is it planned to have two way communication via the bridge? 
As postMessage is a mechanism to post a message from an integration to the host app, the other way around is currently not handled, right?
When we add an integration to the Flutter app we might need a mechanism to tell the integration that it is now hosted within Flutter to e.g. hide certain parts of the UI. Depending on the setup queryParams, event listeners or a global window method come to my mind as possible mechanisms. Let's talk about this at 1 :)

packages/bridge/README.md
```diff
@@ -0,0 +1,142 @@
+# Flip Bridge
+
+The Flip Bridge handles communication between your embeddable apps and the Flip
+App or Flip Admin Console.
+
+**Contents**
```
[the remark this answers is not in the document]
Sqrrl: Yes. As a consumer you should only use the documented functions. The messaging functions are more low level stuff used internally.

packages/bridge/src/messaging/messaging.ts
Sqrrl: Sorry, should have documented that. The `makeRequest` function already listens for the response. In fact, every request should get a response from the host or flutter app. I did a possible implementation in the Console app bridge adapter: https://gitlab.com/flipcorp/flipnext/flip/-/blob/12576-create-basis-for-js-bridge/console/libs/shared/host/src/lib/bridge-adapter/bridge-adapter.service.ts

packages/bridge/src/navigation/navigation.ts
```diff
@@ -0,0 +1,14 @@
+import { v4 as uuidv4 } from "uuid";
+import { makeRequest } from "../messaging";
+import { BridgeMethod } from "../types";
+import { NavigateResult, NavigateRequest } from "./navigation.types";
+
+export function navigate(path: string) {
```
FabianMetzger, line 6: I think for now the path is fine. There might be additional information (e.g. ids) one can send via url-parameters.

packages/bridge/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+**Before using the provided functions, you have to call the `initFlipBridge`
+function to set up the Flip Bridge.**
+
+```js
+import { initFlipBridge } from "@getflip/bridge";
+
+initFlipBridge({
+  debug: true,
+  hostAppOrigin: "http://localhost:4200", // has to be the origin of the targetted host app
```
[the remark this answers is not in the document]
Sqrrl: Fiexed.

packages/bridge/src/navigation/navigation.ts
Sqrrl: Would you prefer a separate parameter to pass url params? Otherwise you could just pass something like `/my-app?param=123` as the path.

FabianMetzger: That's what I thought. For now I do not see a need for extra params but maybe in the future.

webspaceadam (approved):

PR #120, opened 2022-12-09
App 12780 implement new swirl tokens

webspaceadam:
## Summary

[APP-12780](https://flipapp.myjetbrains.com/youtrack/issue/APP-12780)


Sqrrl (commented):
As far as I know, the core tokens are not to be used directly and therefore are not to be documented. Better check with Lenni again. E.g. `core-decorative-light-chilli-surface`

webspaceadam:
core is not documented anymore 🙃

Sqrrl:
nitpick: Would be great to prevent the break here.

Bildschirm­foto 2022-12-14 um 11 35 27

PR #121, opened 2022-12-12
APP-12838 export light and dark theme for tailwind tokens

webspaceadam:
- Ticket: [APP-12838](https://flipapp.myjetbrains.com/youtrack/issue/APP-12838)

### DIFF
--- a/packages/swirl-tokens/scripts/build.ts
+++ b/packages/swirl-tokens/scripts/build.ts
@@ -23,11 +23,7 @@ StyleDictionary.registerFormat({
           mappedTokens[mappedType as keyof mappedTokensType] = {};
         }
 
-        if (theme === "light") {
-          mappedTokens[mappedType as keyof mappedTokensType][name] = value;
-        } else {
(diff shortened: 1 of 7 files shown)


webspaceadam:
The implementation seems a bit odd as we don't extend the tokengroups but i didn't want to produce a breaking change 🌞

packages/swirl-tokens/scripts/build.ts
```diff
@@ -24,14 +24,9 @@ StyleDictionary.registerFormat({
         }
 
         if (theme === "light") {
```
Sqrrl, line 26: **issue:** The if statement seems obsolete.

Sqrrl (changes requested):

packages/swirl-tokens/scripts/build.ts
webspaceadam: yeah it is now

Sqrrl (approved):

PR #128, opened 2022-12-15
App 12906: Add sitemap.xml and robots.txt

webspaceadam:
Ticket: [APP-12906](https://flipapp.myjetbrains.com/youtrack/issue/APP-12906)

### DIFF
--- a/apps/swirl-docs/src/pages/sitemap.xml.ts
+++ b/apps/swirl-docs/src/pages/sitemap.xml.ts
@@ -38,11 +38,9 @@ export async function getServerSideProps({ res }: any) {
 
   const filteredPaths = paths.filter((path) => !path.includes("foundations"));
 
-  // We generate the XML sitemap with the posts data
   const sitemap = generateSiteMap(filteredPaths);
 
   res.setHeader("Content-Type", "text/xml");
-  // we send the XML to the browser
   res.write(sitemap);
   res.end();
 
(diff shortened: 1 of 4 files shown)


apps/swirl-docs/src/pages/sitemap.xml.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  return `<?xml version="1.0" encoding="UTF-8"?>
+   <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+     ${paths
+       .map((path) => {
+         return `
+       <url>
+           <loc>${`${baseUrl}${path}`}</loc>
+           <changefreq>weekly</changefreq>
+           <priority>0.7</priority>
```
Sqrrl, line 14: **question:** Can this be removed if all pages get the same priority anyway?

webspaceadam: we could. i looked on different comparable sites and a few had 0.7 prio on their sites as well. then i googled a bit and found this [article](https://www.contentpowered.com/blog/xml-sitemap-priority-changefreq/#sitemap_priority_attribute) that explained the values in a little bit more depth, but as this param is optional i am totally fine with removing it

Sqrrl: That's fine. Thanks.

Sqrrl (approved):

PR #129, opened 2022-12-15
APP-12902 add bridge dialog methods and events

Sqrrl:
- added new methods and events to README
- implemented methods to create, open, close and destroy dialogs
- implement events to listen to dialog action clicks


webspaceadam (approved):

konnic (approved):
Looks good! 
I've also implemented rudimentary dialog handling in Flutter. From my PoV it would be nice if you could add dialog handling to the demo iframe as well. Nevertheless, that's out of scope for this ticket.

PR #130, opened 2022-12-16
APP-12918 - GTM Integration

webspaceadam:
## Summary

- [Ticket](https://flipapp.myjetbrains.com/youtrack/issue/APP-12918)

### DIFF
--- a/apps/swirl-docs/src/lib/gtm/index.tsx
+++ b/apps/swirl-docs/src/lib/gtm/index.tsx
@@ -1,7 +1,10 @@
 export const GTM_ID = "GTM-TNKB82Q";
 
-// here to get rid of misleading typescript errors
-declare const window: any;
+declare global {
+  interface Window {
+    dataLayer: Record<string, any>;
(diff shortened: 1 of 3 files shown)


apps/swirl-docs/src/lib/gtm/index.tsx
```diff
@@ -0,0 +1,11 @@
+export const GTM_ID = "GTM-TNKB82Q";
+
+// here to get rid of misleading typescript errors
+declare const window: any;
```
Sqrrl, line 4: **suggestion:** 
```ts
declare global {
  interface Window {
    dataLayer: Record<string, unknown>;
  }
}
```

Sqrrl (approved):

apps/swirl-docs/src/lib/gtm/index.tsx
webspaceadam: unknown still gives errors, but i added any instead!

PR #135, opened 2023-01-04
APP-13004 Header Navigation

webspaceadam:
- [Ticket](https://flipapp.myjetbrains.com/youtrack/issue/APP-13004/Header-Navigation-Design-Improvements)


apps/swirl-docs/src/components/HeaderNavigation/index.tsx
```diff
@@ -1,14 +1,42 @@
...
 
 const HeaderNavigation = () => {
-  const [isMobileNavOpen, setIsMobileNavOpen] = useState(false);
   const router = useRouter();
+  const [isMobileNavOpen, setIsMobileNavOpen] = useState(false);
+
+  useEffect(() => {
+    if (typeof window) {
+      if (isMobileNavOpen) {
+        document.getElementsByTagName("body")[0].style.overflow = "hidden";
```
Sqrrl, line 34: **suggestion:** You could consider using https://github.com/willmcpo/body-scroll-lock here. It solves all the edge cases on mobile devices.

apps/swirl-docs/src/components/HeaderNavigation/index.tsx
```diff
@@ -76,21 +76,51 @@ const HeaderNavigation = () => {
...
+      </DesktopView>
+      <MobileView>
+        <header className="sticky top-0 flex flex-col w-full h-full max-h-screen bg-background-default">
+          <div
+            aria-label="main"
+            className="flex justify-between items-center h-16 w-full px-4 border-b-1 font-normal text-base"
+          >
+            <HeaderLogo />
+            {isMobileNavOpen ? (
+              <button type="button" onClick={() => setIsMobileNavOpen(false)}>
```
Sqrrl, line 102: **suggestion:** For better accessibility, the toggle should be inside the navigations `nav` tags (or set `aria-controls` on the button, if not possible). You could also set `aria-expanded="true|false"`. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#menu

apps/swirl-docs/src/components/HeaderNavigation/index.tsx
```diff
@@ -76,21 +76,51 @@ const HeaderNavigation = () => {
...
+                <Image alt="Close Menu" src={icon.src} width={24} height={24} />
+              </button>
+            ) : (
+              <button
```
Sqrrl, line 106: **suggestion:** see above

apps/swirl-docs/src/components/HeaderNavigation/index.tsx
webspaceadam: Had no Idea a lib for that existed. I will take a look into it and implement it :)

Sqrrl (approved):

PR #148, opened 2023-01-19
feat(documentation): :art: layout design improvments

webspaceadam:
## Summary

[APP-13006](https://flipapp.myjetbrains.com/youtrack/issue/APP-13006) Layout Design Improvements

### DIFF
--- a/apps/swirl-docs/src/components/Navigation/LinkedHeaders.tsx
+++ b/apps/swirl-docs/src/components/Navigation/LinkedHeaders.tsx
@@ -14,6 +14,7 @@ const H1 = ({ id, ...rest }: HeadlineProps) => {
 };
 
 const H2 = ({ id, ...rest }: HeadlineProps) => {
+  console.log("hello world");
   if (id) {
     return <h2 id={id} {...rest} />;
   }
--- a/apps/swirl-docs/src/styles/globals.css
+++ b/apps/swirl-docs/src/styles/globals.css
@@ -48,15 +48,19 @@ body {
 }
 
 h1 {
-  font-size: var(--s-font-size-2xl);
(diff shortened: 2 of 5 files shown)


apps/swirl-docs/src/components/Navigation/LinkedHeaders.tsx
```diff
@@ -14,6 +14,7 @@ const H1 = ({ id, ...rest }: HeadlineProps) => {
 };
 
 const H2 = ({ id, ...rest }: HeadlineProps) => {
+  console.log("hello world");
```
Sqrrl, line 17: **chore:** 😁

apps/swirl-docs/src/styles/globals.css
```diff
@@ -48,15 +48,19 @@ body {
 }
 
 h1 {
-  font-size: var(--s-font-size-2xl);
+  font-size: 36px;
```
Sqrrl, line 51: **issue:** Let's use relative units to not break browser zoom / font size settings. Applies to the other pixel values as well.

Sqrrl (changes requested):

apps/swirl-docs/src/components/Navigation/LinkedHeaders.tsx
webspaceadam: hoppla 😅

apps/swirl-docs/src/styles/globals.css
webspaceadam: yeah good point. my fault. started with rem and then changed to px unnecessarilily

Sqrrl (approved):

PR #205, opened 2023-03-14
App 14113 extend codepreview

webspaceadam:
## Summary

[Ticket](https://flipapp.myjetbrains.com/issue/APP-14113) Prepare getflip.dev for API Docs)

### requirements
- be able to display HTTP-Calls in the CodePreview Component
- Be able to sprinkle in different parts of "main" content into the Documentation Layout

### Why the Documentation Layout needs a Refactor: 
The Documentation Layout is a reusable component that is used across all (currently 3 domains: components, icons, tokens) and will be used for every domain that will be implemented and could be implemented in the future. Currently it is controlled through a couple of props, but going above 2-3 cases (e.g. should we show a component preview? is this a component site? should we display the ToC?, etc.) that can be handled through props (e.g. `[tokenDoc]` passes different props than `[componentDoc]`) becomes hardly maintainable over time. Thats why this is a good time to refactor the `DocumentationLayout` to use its own `DocumentationLayoutContext` and make it compoundable. This has another nice Side-Effect: we can reuse it for the `index` pages. 

The coming API for the `DocumentationLayout`: 
```tsx
<DocumentationLayout 
    data={...used as context data...}
    header={...}
    content={...}
    footer={...}
/>
```

with this API each Page is more maintainable through the component composition instead of propdrilling. From `config` to `lego`.


apps/swirl-docs/src/components/CodePreview/CodePreview.tsx
```diff
@@ -0,0 +1,122 @@
...
+ * Let's you easily render syntax highlighted code.
+ * Children of the Component will be displayed in the Header.
+ */
+export function CodePreview({ children, codeExample }: CodePreviewProps) {
+  const [isExpanded, setIsExpanded] = useState<boolean>(false);
+  const [isCopied, setIsCopied] = useState<boolean>(false);
+
+  return (
+    <NoSsr>
+      <CodePreviewContext.Provider value={codeExample}>
```
Sqrrl, line 34: **question:** Is this really necessary? Looking at the usages of `useCodePreviewContext` it seems like you could achieve the same by passing the the code example via props.

apps/swirl-docs/src/components/CodePreview/CodePreview.tsx
```diff
@@ -0,0 +1,122 @@
...
+            <CopyToClipboard
+              text={codeExample.code}
+              onCopy={() => {
+                setIsCopied(true);
+                setTimeout(() => {
+                  setIsCopied(false);
+                }, 2000);
+              }}
+            >
+              <button className="flex justify-center items-center text-[#F2F2F2] text-base font-medium">
```
Sqrrl, line 60: **suggestion:** `type="button"`

apps/swirl-docs/src/components/CodePreview/CodePreview.tsx
```diff
@@ -0,0 +1,122 @@
...
+                {isCopied ? "copied!" : "copy"}
```
Sqrrl, line 61: **nitpick:** "copy" is a bit weak as a label. I think "Copy code" and "Code copied." would be more concise.

apps/swirl-docs/src/components/CodePreview/CodePreview.tsx
```diff
@@ -0,0 +1,122 @@
...
+              <pre
+                className={classNames(
+                  "cursor-text overflow-auto pt-space-16 md:pt-space-8 px-space-24 ",
+                  "md:pb-16",
+                  {
+                    "pb-space-16": !isExpanded,
+                    "pb-16": isExpanded,
+                  }
+                )}
+              >
```
Sqrrl, line 87: **suggestion:** Can we wrap the lines with a `code` element? 

```html
<pre>
  <code>
    …
  </code>
</pre>
```

apps/swirl-docs/src/components/CodePreview/CodePreviewRequestString.tsx
```diff
@@ -0,0 +1,18 @@
+import { useCodePreviewContext } from "./CodePreviewContext";
+
+export function Request() {
```
Sqrrl, line 3: **suggestion:** The naming is a bit generic. Maybe something like "APIEndpointHeader"?

apps/swirl-docs/src/components/Documentation/DocumentationHeader.tsx
```diff
@@ -1,20 +1,15 @@
...
-> = ({ frontMatter }) => {
   return (
     <header className="border-b-1 pb-12 mb-12">
       <div className="mb-space-8 inline-flex items-center">
         <h1 className="text-4xl text-text-default font-font-weight-bold">
-          {frontMatter.title}
+          {frontMatter?.title}
         </h1>
-        {frontMatter.tags?.map((tag) => (
+        {frontMatter?.tags?.map((tag: any) => (
```
Sqrrl, line 12: **suggestion:** `:any` -> `:string`

apps/swirl-docs/src/components/Layout/DocumentationLayout.tsx
```diff
@@ -1,62 +1,61 @@
...
+        {data.navigationLinks && <CategoryNav />}
+        <div className="h-full w-full">
+          <main
+            id="main"
+            className={classNames(
+              "grid grid-cols-1 md:grid-cols-[minmax(0,_45rem)_16rem] gap-8 justify-center",
+              "my-0 mx-auto mt-14 mb-4 md:mb-0 px-4 lg:px-0"
+            )}
+          >
+            <article className="w-full max-w-[45rem]">
```
Sqrrl, line 41: **question:** Is it safe to assume that this can always be an article? What if you want to have multiple articles inside. `content`?

Sqrrl (changes requested):

apps/swirl-docs/src/components/CodePreview/CodePreview.tsx
webspaceadam: I think this is the more elegant solution for a composable component like this one. as a user you just decide which header you want to display and share the data once, the rest of the component gets them through the context. With this we have less prop drilling and less interfaces. when we extend the possibilities of the header (for example i totally see the usecase for changing the displayed language (cURL -> node/kotlin/whatever). with this solution it is just one interface that has to change and you can use it right away. basically comes down to less propdrilling 🙂

Sqrrl (approved):

PR #324, opened 2023-07-04
APP-15624 api docs request response schemas

Sqrrl:
https://getflipdev-qyjze02xj-getflip.vercel.app/

### DIFF
--- a/apps/swirl-docs/src/components/Documentation/Parameter.tsx
+++ b/apps/swirl-docs/src/components/Documentation/Parameter.tsx
@@ -2,8 +2,11 @@ import { Tag } from "../Tags";
 import { ReactMarkdown } from "react-markdown/lib/react-markdown";
 import { Text } from "../swirl-recreations";
 import classNames from "classnames";
+import { ReactNode, useState } from "react";
+import { SwirlIconAdd, SwirlIconRemove } from "@getflip/swirl-components-react";
 
 interface ParameterProps {
+  children?: ReactNode;
   name: string;
   type: string;
   required?: boolean;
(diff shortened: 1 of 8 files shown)


apps/swirl-docs/src/components/Documentation/Parameter.tsx
```diff
@@ -2,33 +2,54 @@ import { Tag } from "../Tags";
...
         "only:rounded-border-radius-sm only:border-border-1"
       )}
     >
-      <div className="flex items-center mb-2 only:mb-0">
-        <code className="text-sm font-font-weight-bold mr-2">{name}</code>
-        <Tag content={type} />
+      <div className="flex items-center">
+        <code
+          className={classNames("text-sm font-font-weight-bold mr-2", {
+            "text-text-success": name.match(/^2\d\d$/),
```
webspaceadam, line 44: **praise:** love the solution

apps/swirl-docs/src/components/Documentation/Parameter.tsx
```diff
@@ -2,33 +2,54 @@ import { Tag } from "../Tags";
...
 
 export function Parameter({
+  children,
   name,
   type,
   description,
   required,
+  enumValues,
 }: ParameterProps) {
+  const [expanded, setExpanded] = useState(false);
```
webspaceadam, line 25: **nitpick:** in most components i try to handle boolean variables with `is` or `has` prefixes.

so here isExpanded and setIsExpanded

apps/swirl-docs/src/components/Documentation/Parameter.tsx
```diff
@@ -42,11 +63,61 @@ export function Parameter({
...
             ),
           }}
-          className="text-sm text-text-default"
+          className="text-sm text-text-default mt-2"
         >
           {description}
         </ReactMarkdown>
       )}
+      {enumValues?.length && (
+        <div className="mt-space-8">
```
webspaceadam, line 72: **nitpick (non-blocking):** i implemented framer motion as a animation lib already (see sidebar navigation / category navigation) for reference. For the UX i think it would be a cool addition to also see it here. should not be that big of an effort :)

webspaceadam (changes requested):
looks great! just some tiny considerations :)

apps/swirl-docs/src/components/Documentation/Parameter.tsx
Sqrrl: Done.

apps/swirl-docs/src/components/Documentation/Parameter.tsx
Sqrrl: Done.

PR #485, opened 2023-12-12
APP-20227 fix mobile icon drawer

Sqrrl:


webspaceadam:
Seems it still has some issues. It does not close immediatly every time, but still i see things getting closed :D


https://github.com/getflip/swirl/assets/23700195/84f7b806-ed51-4824-9d73-2d725a4a831e

Sqrrl:
> Seems it still has some issues. It does not close immediatly every time, but still i see things getting closed :D
> 
>  Bildschirmaufnahme.2023-12-12.um.15.33.44.mov

Should be working better now. There is still a compatibility issue with React where the trigger element of the popover got detached from the DOM while closing. Good enough for now, I hope.

apps/swirl-docs/src/components/Icons/IconGridItem.tsx
```diff
@@ -75,22 +75,24 @@ const IconGridItem: FunctionComponent<IconGridProps> = ({
...
-            onClick={() => {
-              handleTileClick(icons[icon]?.name);
-            }}
-          >
-            <i
-              className={`swirl-icons-${icons[icon]?.name}28 text-icon-strong`}
-            ></i>
-            <IconName />
-          </a>
+          <SwirlPopoverTrigger popover="icon-popover">
```
webspaceadam, line 78: **issue:** looks good now when it comes to opening the tile. But it seems we still have the update issue

https://github.com/getflip/swirl/assets/23700195/f7b9c84c-0f6e-4f37-a0d9-3eca53c7f45a

Sqrrl: Works fine on my machine. Did you run it locally? If so, you have to run the swirl component builds.

webspaceadam: I only made a clean install. forgot i had to build it. works!

webspaceadam (approved):

PR #491, opened 2023-12-14
generate example request body

fellmann:


apps/swirl-docs/src/lib/docs/src/oasBuilder.ts
```diff
@@ -332,45 +336,25 @@ export default class OASBuilder implements IOASBuilder {
...
-        objectivec: String(
-          oasToSnippet(this.oas, operation, {}, {}, "objectivec").code
-        ),
-        powershell: String(
-          oasToSnippet(this.oas, operation, {}, {}, "powershell").code
-        ),
-        r: String(oasToSnippet(this.oas, operation, {}, {}, "r").code),
-      },
+      snippets: Object.fromEntries(
+        (Object.keys(supportedLanguages) as SupportedTargets[]).map((l) => [
```
webspaceadam, line 343: **nitpick (blocking):** we currently don't have documented coding styles like this, but i prefer using readable names for variables. so l should be language here :)

apps/swirl-docs/src/lib/docs/src/oasBuilder.ts
```diff
@@ -332,45 +336,29 @@ export default class OASBuilder implements IOASBuilder {
   } {
     const har = OASToHar(this.oas, operation);
     const harRequest = har.log.entries[0].request;
+    const body = operation.getRequestBodyExamples()[0]?.examples[0]?.value;
+
+    const allLanguages = Object.keys(supportedLanguages) as SupportedTargets[];
+    const allLanguageSnippets = allLanguages.map((language) => [
```
webspaceadam, line 342: **praise:** +1 for this :)

webspaceadam (approved):

PR #498, opened 2023-12-19
APP-20430 Implement new endpoint method tag design

fellmann:


webspaceadam (approved):
big relieve to say goodbye to this useEffect

PR #501, opened 2023-12-20
APP-20302 Design Mobile Navigation View

fellmann:


apps/swirl-docs/src/lib/navigation/src/data/navItems.ts
```diff
@@ -9,21 +9,57 @@ export const navItems: NavItem[] = [
...
+        isExternal: true,
+      },
+      {
+        title: "Public Users- and Groups-Sync-API",
+        url: "https://base.flip-app.com/openapi/external/sync",
+        isExternal: true,
+      },
+    ],
+  },
+  {
```
webspaceadam, line 47: issue: as we currently have no content for those pages i would not show them at all :)

apps/swirl-docs/src/lib/docs/src/oasBuilder.ts
```diff
@@ -189,10 +190,7 @@ export default class OASBuilder implements IOASBuilder {
               OASBuilder.X_FLIP_INTERNAL
             );
 
-            if (
-              isInternal &&
-              process.env.NEXT_PUBLIC_DEPLOYMENT_STAGE !== "staging"
-            ) {
+            if (isInternal && isProdDeployment) {
```
webspaceadam, line 193: +1 for the cleanup and usage of the constant :)

apps/swirl-docs/src/lib/docs/src/MergedYamlOperations.ts
```diff
@@ -39,7 +39,7 @@ function isYamlFile(file: string) {
 function createSpecPath(spec: string) {
```
webspaceadam, line 39: as this function is the same as in multipleDocs i think we can share the functionality here and export one of them and import in the other file when needed

apps/swirl-docs/src/components/Navigation/SidebarNavigation/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  );
+
+  if (navItemBase?.devOnly && isProdDeployment) {
+    return null;
+  }
+
+  const navChildren = navItemBase?.children;
+
+  if (!navChildren?.length) {
+    return null;
```
webspaceadam, line 26: please use undefined. null is an object in javascript and not a "real" null value.

webspaceadam: image
vs 

image

apps/swirl-docs/src/components/Navigation/SidebarNavigation/SidebarNavItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+import Link from "next/link";
+import { HttpMethods } from "oas/dist/rmoas.types";
+import { forwardRef, useRef } from "react";
+import {
+  Tag,
+  mapHttpMethodToTagContent,
+  mapHttpMethodToTagScheme,
+} from "../../Tags";
+
+export const SidebarNavItem = forwardRef<
```
webspaceadam, line 12: As those Nav Items are now used for both the mobile nav and the sidebar nav, the sidebar nav naming here is misleading. we should call them navigation items

apps/swirl-docs/src/components/Navigation/SidebarNavigation/SidebarNavSubItem.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { NavItem } from "@swirl/lib/navigation/";
+import classNames from "classnames";
+import { AnimatePresence, motion } from "framer-motion";
+import Image from "next/image";
+import Link from "next/link";
+import { useEffect, useState } from "react";
+import { ANIMATION_CLOSED, ANIMATION_OPENED } from ".";
+import { SidebarNavItem } from "./SidebarNavItem";
+
+export const SidebarNavSubItem = ({
```
webspaceadam, line 11: I think the structure makes it hard to read and understand which navItem is rendered when. 

The ternaries for the conditional rendering are just too deeply nested. 

A few weeks ago i tidied up the mobile navigation to use a strategy pattern for recursevly rendering the navigation and i think this strategy can be also applied here. They were in the `HeaderNavigation/Mobile` package. 

There are several benefits for doing that:
- Makes it easier to understand what the current component is (they are all sidebarnavsubitems atm)
- makes the code more maintainable and extendable in the future

apps/swirl-docs/src/pages/sitemap.xml.ts
```diff
@@ -23,17 +23,21 @@ function SiteMap() {
   // getServerSideProps will do the heavy lifting
 }
 
-export async function getServerSideProps({ res }: any) {
-  const paths: string[] = [];
-
+function addNavItemsToPaths(paths: string[], navItems: NavItem[]) {
   navItems.forEach((item) => {
```
webspaceadam, line 27: i know forEach modifies the passed in array, but i do not like this implicit behaviour. i think it would be better to use map and use a return value. in that way we can directly call this function in line 38 and just return the array directly instead of creating the const and manipulating it through reference

webspaceadam (changes requested):
I really like, that you merged the mobile and sidebar nav. 

But i think we should change the structure of how it is determined which markup is used for which navitem. Right now it is quite confusing.

apps/swirl-docs/src/lib/docs/src/MergedYamlOperations.ts
fellmann: removed unused code

apps/swirl-docs/src/components/Navigation/SidebarNavigation/SidebarNavSubItem.tsx
fellmann: restructured navigation, please have a look

apps/swirl-docs/src/components/Navigation/SidebarNavigation/index.tsx
fellmann: I don't get this to work with TypeScript. 
```
'SidebarNavigation' cannot be used as a JSX component.
  Its return type 'Element | null | undefined' is not a valid JSX element.
    Type 'undefined' is not assignable to type 'Element | null'.ts(2786)
```

webspaceadam: alright. then we'll go with null

webspaceadam (approved):
lgtm :)

PR #506, opened 2024-01-05
APP-20517 Refactor EndpointDescription into smaller components

fellmann:


webspaceadam (approved):
after APP-20616 was merged this should be updated otherwise lgtm :)

PR #507, opened 2024-01-05
APP-20616 Fix and unify markdown and mdx formatting

fellmann:


webspaceadam (approved):
i really like the approach and makes everything a lot cleaner :)

PR #512, opened 2024-01-10
App 20189 add step to integrate api docs to algolia index

webspaceadam:
## Summary

- postbuild script refactored
- created algoliarecord generators to have an implementable interface for the future and have a maintainable design
- implemented vitest and tests for the generators to not push garbage to algolia


packages/swirl-tokens/dart/lib/styles.dark.dart
```diff
@@ -4,7 +4,7 @@
 //
 
 // Do not edit directly
-// Generated on Fri, 15 Dec 2023 15:44:18 GMT
```
fellmann, line 7: **question:** this is a bit annoying ... can the timestamp be removed somehow?

packages/swirl-icons/dart/lib/fonts/swirl-icons.ttf
fellmann, line 1: **question:** this is a bit annoying ... why do these file change?

apps/swirl-docs/tests/scripts/AlgoliaDataGenerator/AlgoliaDataFactory.test.ts
fellmann, line 1: **question:** do we want to keep tests in a separate directory, or next to the source files?

fellmann (approved):
lgtm

packages/swirl-tokens/dart/lib/styles.dark.dart
webspaceadam: my current state is that this is not possible unfortunately. but the answer for why maybe has @Sqrrl

packages/swirl-icons/dart/lib/fonts/swirl-icons.ttf
webspaceadam: part of the build / dev step

apps/swirl-docs/tests/scripts/AlgoliaDataGenerator/AlgoliaDataFactory.test.ts
webspaceadam: Yeah i thought about that. But as the whole app will grow i thought having them in a seperate directory keeps the source code "clean". but no hard feelings here. we can also change that :) what is your take on that?

PR #520, opened 2024-01-12
[APP-20863] [APP-20864] sort resource, improve path param in code sippet

fellmann:
additionally, add endpoint overview


webspaceadam (approved):
lgtm! i have to say i really like the overview. additional idea: maybe we will add a description for the resource to use the space below the headline?

PR #603, opened 2024-02-22
Fix Storybook form input props

joseabernardes:
## Summary

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### General

- [x] The changes do not contain any breaking changes.
- [x] The changes do not introduce new components that don't belong in the library (e.g. non-reusable components, highly specialized components, components including business logic)
  - [ ] All tests are 🟢.
  - [ ] No static text is used.

### DIFF
--- a/packages/swirl-components/src/components/swirl-color-input/swirl-color-input.stories.ts
+++ b/packages/swirl-components/src/components/swirl-color-input/swirl-color-input.stories.ts
@@ -1,4 +1,4 @@
-import { generateStoryElement } from "../../utils";
+import { addArgumentsToElement, generateStoryElement } from "../../utils";
 import Docs from "./swirl-color-input.mdx";
 
 export default {
@@ -16,8 +16,7 @@ const Template = (args) => {
   const formControl = document.createElement("swirl-form-control");
   const element = generateStoryElement("swirl-color-input", args);
 
-  formControl.setAttribute("label", "Color input");
-  formControl.setAttribute("disabled", args.disabled ? "true" : "false");
+  addArgumentsToElement(formControl, { ...args, label: "Color input" });
   formControl.append("\n  ", element, "\n");
 
   return formControl;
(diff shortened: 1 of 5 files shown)


joseabernardes:
Does it make sense to add a new changelog entry in this case? I'm asking because the PR only changes storybook files, and not actual lib code.

packages/swirl-components/src/components/swirl-color-input/swirl-color-input.stories.ts
```diff
@@ -16,8 +16,7 @@ const Template = (args) => {
   const formControl = document.createElement("swirl-form-control");
   const element = generateStoryElement("swirl-color-input", args);
 
-  formControl.setAttribute("label", "Color input");
-  formControl.setAttribute("disabled", args.disabled ? "true" : "false");
+  addArgumentsToElement(formControl, { ...args, label: "Color input" });
```
joseabernardes, line 19: I was a bit undecided about the approach to take here...

Initially I thought about creating a helper function in which I specifically added the attributes (description, disabled, inline, invalid, label ), but then I thought that if in the future we add any more attributes to the form-control, we would have to remember to update the function.

So I thought of this solution of injecting all the attributes of the input into the form-control. The disadvantage of this solution is that we'll be adding attributes we don't need to the form-control.
In your opinion, does that feel unsafe? Should I go with the more declarative approach?

Sqrrl:
> Does it make sense to add a new changelog entry in this case? I'm asking because the PR only changes storybook files, and not actual lib code.

No. No changeset necessary.

packages/swirl-components/src/components/swirl-color-input/swirl-color-input.stories.ts
Sqrrl: `args` only contains custom props that should always be distinct from native HTML attributes. So adding them to components that might not support them should not be an issue.

Sqrrl, line 19: Suggestion: `addArgumentsToElement`-> `addAttribtuesToElement`

joseabernardes: Sounds good!

Sqrrl (approved):

PR #607, opened 2024-02-23
Add option to open the datepicker at focus in swirl-date-input

joseabernardes:
## Summary

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### General

- [x] The submitted code is organized and formatted according to our [💅 Style Guide](https://swirl-storybook.flip-app.dev/?path=/docs/requirements-style-guide--docs).

  - [ ] All tests are 🟢.
  - [ ] No static text is used.

### DIFF
--- a/packages/swirl-components/src/components/swirl-date-input/swirl-date-input.spec.tsx
+++ b/packages/swirl-components/src/components/swirl-date-input/swirl-date-input.spec.tsx
@@ -4,9 +4,16 @@ jest.mock("maska/dist/es6/maska", () => ({
   create: maskSpy,
 }));
 
+(global as any).IntersectionObserver = class {
+  constructor() {}
+  disconnect() {}
+  observe() {}
+};
+
 import { newSpecPage } from "@stencil/core/testing";
 
 import { SwirlDateInput } from "./swirl-date-input";
+import { SwirlPopover } from "../swirl-popover/swirl-popover";
 
 describe("swirl-date-input", () => {
   beforeEach(() => {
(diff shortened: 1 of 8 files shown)


packages/swirl-components/src/components/swirl-date-input/swirl-date-input.tsx
```diff
@@ -46,6 +46,7 @@ export class SwirlDateInput {
   @Prop() labels?: WCDatepickerLabels;
   @Prop() locale?: string = "en-US";
   @Prop() placeholder?: string = "yyyy-mm-dd";
+  @Prop() preferredInputMode?: "input" | "pick" = "pick";
```
Sqrrl, line 49: Issue: To not make this a breaking change, the default should be "input".

Sqrrl (changes requested):
Issue: When I tab into the form control, the picker opens correctly. When I tab again to interact with the picker, the picker closes.

packages/swirl-components/src/components/swirl-date-input/swirl-date-input.tsx
joseabernardes: Sure, it was just to help me test on mobile :P

joseabernardes:
> Issue: When I tab into the form control, the picker opens correctly. When I tab again to interact with the picker, the picker closes.

Excellent point!
I was investigating, and it appears to be an already existing behavior, and the reason for it is because when we give focus to the popover(datepicker) via tab, the popover-trigger(calendar button) gets hidden (because the input itself loses focus) and then the trigger closes the datepicker when we try to navigate in it (via tab or mouse).
One of the solutions that I found was to set the `hidePopoverWhenInvisible` prop to false on the trigger, but that could have unwanted side effects. 
What do you think about this approach?

Sqrrl:
> > Issue: When I tab into the form control, the picker opens correctly. When I tab again to interact with the picker, the picker closes.
> 
> Excellent point! I was investigating, and it appears to be an already existing behavior, and the reason for it is because when we give focus to the popover(datepicker) via tab, the popover-trigger(calendar button) gets hidden (because the input itself loses focus) and then the trigger closes the datepicker when we try to navigate in it (via tab or mouse). One of the solutions that I found was to set the `hidePopoverWhenInvisible` prop to false on the trigger, but that could have unwanted side effects. What do you think about this approach?

We could try and keep the form control "open" when the date picker has focus, so the calendar icon doesn't disappear.

packages/swirl-components/src/components/swirl-form-control/swirl-form-control.tsx
```diff
@@ -173,7 +173,7 @@ export class SwirlFormControl {
   private onKeyDown = (event: KeyboardEvent) => {
     if (event.key === "Tab") {
       setTimeout(() => {
-        if (!this.el.contains(getActiveElement())) {
+        if (!isDescendantOf(getActiveElement(), this.el)) {
```
joseabernardes, line 176: The issue with the tab navigation was due to this.
As the popover was inside a Shadow Root, it wasn't considered a descendent of the control itself, so the input would lose focus here.
I then implemented the `isDescendantOf ` utility method which checks whether an element is a descendant of another, even if it is inside a Shadow Root.

packages/swirl-components/src/components/swirl-date-input/swirl-date-input.spec.tsx
```diff
@@ -1,16 +1,34 @@
 const maskSpy = jest.fn();
```
Sqrrl, line 1: Issue: The tests are technically green, but running them locally `yarn stencil:test:watch` produces some errors caused by the setTimeout usage in focus().

PR #623, opened 2024-03-11
feat(swirl): show tooltip on label outside and not mobile

manuel-pross:
## Summary

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### General

  - [ ] All tests are 🟢.
  - [ ] No static text is used.

### DIFF
--- a/packages/swirl-components/src/components/swirl-color-input/swirl-color-input.tsx
+++ b/packages/swirl-components/src/components/swirl-color-input/swirl-color-input.tsx
@@ -35,6 +35,7 @@ export class SwirlColorInput {
   @Prop() pickerLabel?: string = "Color picker";
   @Prop() placeholder?: string;
   @Prop() required?: boolean;
+  @Prop() toolip?: string;
   @Prop({ mutable: true, reflect: true }) value?: string;
 
   @Event() inputBlur: EventEmitter<FocusEvent>;
--- a/packages/swirl-components/src/components/swirl-color-input/swirl-color-input.tsx
+++ b/packages/swirl-components/src/components/swirl-color-input/swirl-color-input.tsx
@@ -35,7 +35,6 @@ export class SwirlColorInput {
   @Prop() pickerLabel?: string = "Color picker";
   @Prop() placeholder?: string;
   @Prop() required?: boolean;
-  @Prop() toolip?: string;
   @Prop({ mutable: true, reflect: true }) value?: string;
 
   @Event() inputBlur: EventEmitter<FocusEvent>;
--- a/packages/swirl-components/src/components/swirl-form-control/swirl-form-control.css
+++ b/packages/swirl-components/src/components/swirl-form-control/swirl-form-control.css
@@ -126,6 +126,21 @@
   @media (--from-desktop-without-touch) {
     --swirl-text-input-placeholder-size: var(--s-font-size-sm);
   }
+
+  & .form-control__label-text {
+    @media (--from-desktop-without-touch) {
+      display: flex;
+      align-items: center;
+    }
+
+    & .form-control__tooltip {
+      display: none;
+      @media (--from-desktop-without-touch) {
+        display: inline-flex;
+        margin-left: 4px;
+      }
+    }
+  }
 }
 
 .form-control--invalid {
--- a/packages/swirl-components/src/components/swirl-form-control/swirl-form-control.css
+++ b/packages/swirl-components/src/components/swirl-form-control/swirl-form-control.css
@@ -137,7 +137,7 @@
       display: none;
       @media (--from-desktop-without-touch) {
         display: inline-flex;
-        margin-left: 4px;
+        margin-left: var(--s-space-4);
       }
     }
   }
--- a/packages/swirl-components/src/components/swirl-form-control/swirl-form-control.css
+++ b/packages/swirl-components/src/components/swirl-form-control/swirl-form-control.css
@@ -136,7 +136,7 @@
     & .form-control__tooltip {
       display: none;
       @media (--from-desktop-without-touch) {
-        display: inline-flex;
+        display: flex;
         margin-left: var(--s-space-4);
       }
     }
--- a/packages/swirl-components/src/components/swirl-form-control/swirl-form-control.css
+++ b/packages/swirl-components/src/components/swirl-form-control/swirl-form-control.css
@@ -138,6 +138,10 @@
       @media (--from-desktop-without-touch) {
         display: flex;
         margin-left: var(--s-space-4);
+
+        &:hover {
+          cursor: default;
+        }
       }
     }
   }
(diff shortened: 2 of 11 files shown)


packages/swirl-components/src/components/swirl-form-control/swirl-form-control.css
```diff
@@ -126,6 +126,21 @@
...
+    @media (--from-desktop-without-touch) {
+      display: flex;
+      align-items: center;
+    }
+
+    & .form-control__tooltip {
+      display: none;
+      @media (--from-desktop-without-touch) {
+        display: inline-flex;
+        margin-left: 4px;
```
Sqrrl, line 140: Issue: Use a design token `4px` -> `var(--s-space-4)`

packages/swirl-components/src/components/swirl-form-control/swirl-form-control.tsx
```diff
@@ -209,6 +209,12 @@ export class SwirlFormControl {
 
     const isSelect = this.inputEl.tagName === "SWIRL-SELECT";
 
+    const tooltip: string | null = (() => {
```
Sqrrl, line 212: Question: Why the IIFE?

packages/swirl-components/src/components/swirl-color-input/swirl-color-input.tsx
```diff
@@ -35,6 +35,7 @@ export class SwirlColorInput {
   @Prop() pickerLabel?: string = "Color picker";
   @Prop() placeholder?: string;
   @Prop() required?: boolean;
+  @Prop() toolip?: string;
```
Sqrrl, line 38: Issue: The tooltip prop should be part of the form control component. No need to add it to every input component.

packages/swirl-components/src/components/swirl-form-control/swirl-form-control.tsx
Sqrrl, line 212: Issue: I think this should be a prop.

packages/swirl-components/src/components/swirl-form-control/swirl-form-control.tsx
```diff
@@ -242,7 +248,15 @@ export class SwirlFormControl {
             <LabelTag class="form-control__label" onClick={this.onLabelClick}>
               <span class="form-control__label-text" id={this.labelId}>
                 {this.label}
+                {tooltip && (
+                  <span class="form-control__tooltip">
+                    <swirl-tooltip content={tooltip} positioning="fixed">
+                      <swirl-icon-help size={16}></swirl-icon-help>
```
Sqrrl, line 254: Issue: The icon should be focusable to be able to open the tooltip via keyboard. (`tabindex="0"`)

packages/swirl-components/src/components/swirl-form-control/swirl-form-control.tsx
manuel-pross: I did not want to use a function which is called in the template so I thought this would be a good choice. Replaced it with:

`const tooltipEl = this.inputEl.getAttribute("tooltip");
const tooltip: string | null = tooltipEl !== null ? tooltipEl : null;`

packages/swirl-components/src/components/swirl-color-input/swirl-color-input.tsx
manuel-pross: That was some legacy code from testing. Thanks for the catch ;)

packages/swirl-components/src/components/swirl-form-control/swirl-form-control.tsx
manuel-pross: Done. This makes the Question regarding the IIFE obsolete right?

Sqrrl: Yes

packages/swirl-components/src/components/swirl-color-input/swirl-color-input.tsx
manuel-pross: Yes you are right

manuel-pross (commented):
The requirement is that the tooltip only works on desktop (where the hover exists) and only when the label position is 'outside'. 

I have implemented this in the code with an additional check for the label position in the template. Is there also a good way to represent this in Storybook?

manuel-pross (commented):
Done

packages/swirl-components/src/components/swirl-form-control/swirl-form-control.tsx
```diff
@@ -242,7 +243,15 @@ export class SwirlFormControl {
             <LabelTag class="form-control__label" onClick={this.onLabelClick}>
               <span class="form-control__label-text" id={this.labelId}>
                 {this.label}
+                {this.tooltip && this.labelPosition === "outside" && (
+                  <span class="form-control__tooltip">
+                    <swirl-tooltip content={this.tooltip} positioning="fixed">
```
webspaceadam, line 248: **issue:** 

i dont know how long the texts in the tooltip will get, but we should ensure it renders correctly. i just checked if the tooltip component itself is able to display a very long text
image it does work. 

unfortunately the same or just a lengthy text gets cut in the component here. 
image

webspaceadam (changes requested):

packages/swirl-components/src/components/swirl-select/swirl-select.tsx
```diff
@@ -41,6 +41,7 @@ export class SwirlSelect implements SwirlFormInput<string[]> {
   @Prop() searchPlaceholder?: string;
   @Prop() selectId?: string = Math.round(Math.random() * 1000000).toString();
   @Prop() swirlAriaDescribedby?: string;
+  @Prop() tooltip?: string;
```
ttobollik, line 44: **question:** Why is the prop required here? Isn't it always set on the form-control? Also i cannot see any change in storybook once i set it.

packages/swirl-components/src/components/swirl-form-control/swirl-form-control.css
```diff
@@ -126,6 +126,21 @@
...
   }
+
+  & .form-control__label-text {
+    @media (--from-desktop-without-touch) {
+      display: flex;
+      align-items: center;
+    }
+
+    & .form-control__tooltip {
+      display: none;
```
ttobollik, line 137: **issue:** I cannot see the tooltip, even when i set it and change the label to 'outside'. it is still set to display: none

ttobollik (changes requested):
You can check my comments for requested changes :)

packages/swirl-components/src/components/swirl-select/swirl-select.tsx
manuel-pross: You are right it is not required. At my first attempt to implement this I added it on every component which could be nested inside a form-control and not on the top level form-control.
I was sloppy by removing this code and missed that one.
Thanks :)

ttobollik (approved):

packages/swirl-components/src/components/swirl-form-control/swirl-form-control.tsx
manuel-pross: Done

manuel-pross (commented):
done

webspaceadam (approved):

PR #696, opened 2024-05-09
feat(bridge): Create showToast method

joseabernardes:
## Summary

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### General

  - [ ] All tests are 🟢.
  - [ ] No static text is used.

### DIFF
--- a/packages/bridge/README.md
+++ b/packages/bridge/README.md
@@ -13,6 +13,7 @@ App or Flip Admin Console.
   - [Theming](#theming)
   - [Dialogs](#dialogs)
   - [Modals](#modals)
+  - [Toasts](#toasts)
 - [Events](#events)
 - [Error Handling](#error-handling)
 - [Development](#development)
@@ -360,6 +361,39 @@ Destroys a modal, removing it from the DOM.
 import { destroyModal } from "@getflip/bridge";
 
 await destroyModal({ id: "my-modal" });
+```
+
+### Toasts
+
+#### `showToast`
+
+Shows a toast by the host app.
+
+**Param**
+
+```js
+{
+  content: string;
+  duration?: number;
+  icon?: string; 
+  intent?: SwirlToastIntent;
+}
+```
+
+**Returns** `Promise<boolean>`
+
+**Example**
+
+```js
+import { showToast } from "@getflip/bridge";
+
+await showToast({
+  content: "My toast",
+  duration: 5000,
+  icon: "icon",
+  intent: "success",
+});
+
 ```
 
 ## Events
--- a/packages/bridge/README.md
+++ b/packages/bridge/README.md
@@ -373,7 +373,7 @@ Shows a toast by the host app.
 
 ```js
 {
-  content: string;
+  text: string;
   duration?: number;
   icon?: string; 
   intent?: SwirlToastIntent;
@@ -388,7 +388,7 @@ Shows a toast by the host app.
 import { showToast } from "@getflip/bridge";
 
 await showToast({
-  content: "My toast",
+  text: "My toast",
   duration: 5000,
   icon: "icon",
   intent: "success",
--- a/packages/bridge/src/toast/toast.types.ts
+++ b/packages/bridge/src/toast/toast.types.ts
@@ -3,7 +3,7 @@ import { BridgeMethod } from "../types";
 import { SwirlToastIntent } from "@getflip/swirl-components/loader";
 
 export type ShowToastRequestParams = {
-  content: string;
+  text: string;
   duration?: number;
   icon?: string;
   intent?: SwirlToastIntent;
--- a/packages/bridge/src/toast/toast.types.ts
+++ b/packages/bridge/src/toast/toast.types.ts
@@ -1,0 +1,17 @@
+import { BridgeRequest } from "../messaging";
+import { BridgeMethod } from "../types";
+import { SwirlToastIntent } from "@getflip/swirl-components/loader";
+
+export type ShowToastRequestParams = {
+  content: string;
+  duration?: number;
+  icon?: string;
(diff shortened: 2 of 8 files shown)


packages/bridge/src/toast/toast.types.ts
```diff
@@ -0,0 +1,17 @@
+import { BridgeRequest } from "../messaging";
+import { BridgeMethod } from "../types";
+import { SwirlToastIntent } from "@getflip/swirl-components/loader";
+
+export type ShowToastRequestParams = {
+  content: string;
+  duration?: number;
```
joseabernardes, line 7: This doesn't fit with the mobile toast duration attribute, because there we have it as an enum `short`, `medium`, `long`.
We need to decide if we drop it for now, or if we make it compatible with both mobile and web.

packages/bridge/src/toast/toast.types.ts
```diff
@@ -0,0 +1,17 @@
...
+  icon?: string;
```
joseabernardes, line 8: This would require a mapper in the mobile app to map from the icon name to the corresponding `SwirlIcons`- `IconData` type.
We need to decide if we do it, or if we drop it for now also.

packages/bridge/README.md
```diff
@@ -362,6 +363,39 @@ import { destroyModal } from "@getflip/bridge";
...
+
+#### `showToast`
+
+Shows a toast by the host app.
+
+**Param**
+
+```js
+{
+  content: string;
```
Sqrrl, line 376: Question: Does the wording "content" imply that HTML is allowed here? I think, otherwise we've used "text" in the past, when only plaintext is allowed.

packages/bridge/src/toast/toast.types.ts
Sqrrl: I think the ms approach is more flexible and explicit than the enum. (What does short mean?) Maybe Flip Mobile can map specific ranges to their enum values?

packages/bridge/src/toast/toast.types.ts
Sqrrl: Maybe we can just use the glyph names of our icons. These should be the same in Swirl and Flip Mobile. Ignoring stuff like camelCase vs kebap-case, of course.

packages/bridge/README.md
joseabernardes: Right, would change it to `text` then to be less ambiguous.

packages/bridge/src/toast/toast.types.ts
joseabernardes: That was my idea, just wanted to raise awareness for the mobile guys that we would maybe require a new mapper/parser in the mobile app, similar to the `flip_legacy_icon_parser`,to map from the glyph names to the `IconData`, because `flip_legacy_icon_parser` only does that for what we call "symbols" in swirl terminology.

packages/bridge/src/toast/toast.types.ts
konnic: We can use the ms approach. On the Dart side we can convert it into a Duration object and drop the enum approach.

packages/bridge/src/toast/toast.types.ts
konnic: Yes we would need something like `flip_legacy_icon_parser` to support icons.

Sqrrl (approved):

packages/bridge/src/toast/toast.types.ts
```diff
@@ -0,0 +1,17 @@
+import { BridgeRequest } from "../messaging";
+import { BridgeMethod } from "../types";
+import { SwirlToastIntent } from "@getflip/swirl-components/loader";
+
+export type ShowToastRequestParams = {
```
konnic, line 5: **question:** Would it make sense to match the [toast component spec](https://www.figma.com/design/iRHiAFsHxnzux8gQ9EjlZe/Swirl-Components?node-id=2-47&t=fdyVnYwTOf4niG6h-0)? 

i.e. supporting:
- text (required)
- type/intent (required)
- leading icon (optional)
- trailing icon (optional)
- trailing text (optional)
- duration (optional)

joseabernardes: TBH I based the spec from the [swirl-toast](https://swirl-storybook.flip-app.dev/?path=/docs/components-swirltoast--docs) spec, because I was not aware of a different one and also because it seemed simple for a first iteration. But we can definitely do it differently.

What your take on this @Sqrrl?

Sqrrl: When using swirl-toast you can achieve leading and trailing icons via the component slot. If the bridge method only uses plaintext, we might need a way to add icons. I'd be fine with leading and trailing icon properties. Not sure what trailing text is supposed to achieve, though.

joseabernardes: I would keep it simple for now and not have the trailing icon and trailing text tbh.
With the `text`, `type/intent`, `icon/leading-icon` and `duration` we can already cover the majority of the use cases for a standard toast, and keeping them consistent with our designs. 
And in the future, we can for sure extend it to have more configuration.

konnic: @Sqrrl trailing text may not be the best term. I was referring to the trailing text button (see image below). But I'm fine with leaving it out, since we do not use it as of now. Just wanted to point out that it exists in the spec.
image

joseabernardes: If you both agree, I would go with this for now, and we improve in the future if we seen the need.

Sqrrl: Sure.

PR #727, opened 2024-06-03
add linger duration to popover hover

AshleyBekemeier:
## Summary

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### General

  - [ ] All tests are 🟢.
  - [ ] No static text is used.


packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
```diff
@@ -14,8 +14,10 @@ export class SwirlPopoverTrigger {
   @Prop() setAriaAttributes?: boolean = true;
   @Prop() swirlPopover!: string | HTMLSwirlPopoverElement;
   @Prop() triggerOnHover?: boolean = false;
+  @Prop() hoverLingerDuration?: number = 200;
```
Sqrrl, line 17: Suggestion: `closingDelay`?

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
```diff
@@ -80,6 +82,8 @@ export class SwirlPopoverTrigger {
 
     popoverEl.open(triggerEl, true);
 
+    this.stopHoverLingerDelay();
```
Sqrrl, line 85: Nitpick: componentDidLoad is only called once at the beginning of the component instance lifecycle. There should be no need to stop the delay handler here.

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
```diff
@@ -99,12 +103,29 @@ export class SwirlPopoverTrigger {
 
   private onMouseleave = () => {
     if (!this.triggerOnHover) return;
-
     const popoverEl = this.getPopoverEl();
 
-    popoverEl.close(true);
+    popoverEl.addEventListener("mouseenter", () => {
```
Sqrrl, line 108: Issue: This will create a new event listener on each mouse leave event. The listeners for "mouseenter" and "mouseleave" should only be added once when a popoverEl is found, probably outside of this onMouseleave method. And you need to make sure to remove the event listeners when the popoverEl is changes or the component is destroyed.

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
Sqrrl, line 17: Issue: A default of 200ms would be a breaking change. Can we use undefined or 0 as the default?

AshleyBekemeier: Why is a default value here a breaking change?

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
AshleyBekemeier: it's called from mouseEnter not in componentDidLoad

Sqrrl: Ah, the code folding of Github confused me.

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
Sqrrl: If you're already using triggerOnHover in your app it will introduce a 200ms closing delay when you update the Swirl version.

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
AshleyBekemeier: Because one popover can have multiple trigger elements i can't have a permanent listener attached. I solved it now by making the event listeners only invoke 'once'.

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
AshleyBekemeier: i don't think anyone was using it anyways but i removed the default :D

AshleyBekemeier:
@Sqrrl After talking with Liu I came to the the conclusion that I had to add a second prop for a hover delay, please review the new functionality as well :)

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
```diff
@@ -98,13 +110,46 @@ export class SwirlPopoverTrigger {
   };
 
   private onMouseleave = () => {
+    clearTimeout(this.hoverDelayReference);
+    // eslint-disable-next-line @stencil/strict-boolean-conditions
```
Sqrrl, line 114: Nitpick: You can prevent this by doing `if (!Boolean(this.hoverDelayReference))`

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
Sqrrl: But a trigger can only have one popover. Multiple triggers adding a mouseenter listener to the popover should be fine.

The once solution still produces listeners that are never removed, unfortunately. For each mouseleave on the trigger you register a mouseenter on the popover. There is no guarantee that the mouseenter on the popover ever happens  and is then unregistered.

AshleyBekemeier: yeah the mouseenter event was fine, but on leave the second trigger element would close the popover. I resolved it now by keeping track of which popover trigger has caused the popover to open, have another look :)

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
```diff
@@ -72,9 +78,34 @@ export class SwirlPopoverTrigger {
...
+    popoverEl.addEventListener("mouseenter", () => {
+      if (this.triggerIsActive) {
+        this.stopHoverLingerTimer();
+      }
+    });
+    popoverEl.addEventListener("mouseleave", () => {
+      if (this.triggerIsActive) {
+        this.mouseleaveHandler();
+      }
+    });
```
Sqrrl, line 93: Just one more clean up step: Web components can implement a `disconnectedCallback()` method to do stuff like remove event handlers when the component is removed from the DOM. These two listeners should be removed there. Otherwise the listeners on the popover stay active even if the trigger has been removed. That would cause errors.

AshleyBekemeier: My understanding is that you cannot remove event handlers that have a reference to the component because they use the arrow functions. Do you know of a way around this?

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_value_of_this_within_the_handler

Sqrrl: You can make the callbacks a class member and use that reference to add and remove the listener. E.g.:

```ts
popoverOnMouseenter = () => {
  if (this.triggerIsActive) {
    this.stopHoverLingerTimer();
  }
}
```

AshleyBekemeier: yeah that's the problem, that doesn't work because the context for 'this' is the element that triggered the event, not the class that defined the callback. very weird javascript stuff.

I made it work now after trial and error by passing in the 'this' context through binding it to the component..

Sqrrl: Using an arrow function for the class member like in my example should work.

`methodName = () => {}` instead of `methodName() {}`

`.bind(this)` works as well, of course.

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
```diff
@@ -41,6 +47,11 @@ export class SwirlPopoverTrigger {
 
   disconnectedCallback() {
     this.intersectionObserver?.disconnect();
+    const popoverEl = this.getPopoverEl();
+    if (Boolean(popoverEl)) {
+      popoverEl.removeEventListener("mouseenter", this.popoverMouseEnter);
+      popoverEl.removeEventListener("mouseenter", this.popoverMouseLeave);
```
Sqrrl, line 53: Issue: Should be "mouseleave"

Sqrrl (approved):

PR #730, opened 2024-06-07
Popover improvements

AshleyBekemeier:
## Summary

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### General

  - [ ] All tests are 🟢.
  - [ ] No static text is used.


Sqrrl (commented):
Issue: Has probably been a problem before your changes, but the trigger always needs to be able to activate the popover via keyboard controls. I would suggest to additionally  trigger the popover via the focus event when triggerOnHover is true.

AshleyBekemeier:
> Issue: Has probably been a problem before your changes, but the trigger always needs to be able to activate the popover via keyboard controls. I would suggest to additionally trigger the popover via the focus event when triggerOnHover is true.

What if we want to implement keyboard support in a different way? should I add a prop that disables keyboard support?

Sqrrl:
> > Issue: Has probably been a problem before your changes, but the trigger always needs to be able to activate the popover via keyboard controls. I would suggest to additionally trigger the popover via the focus event when triggerOnHover is true.
> 
> What if we want to implement keyboard support in a different way? should I add a prop that disables keyboard support?

What do you mean with "different way". From an accessibility perspective focus would be the expected behavior, I think. You could, of course, always implement additional keyboard controls.

AshleyBekemeier:
> > > Issue: Has probably been a problem before your changes, but the trigger always needs to be able to activate the popover via keyboard controls. I would suggest to additionally trigger the popover via the focus event when triggerOnHover is true.
> > 
> > 
> > What if we want to implement keyboard support in a different way? should I add a prop that disables keyboard support?
> 
> What do you mean with "different way"? From an accessibility perspective focus would be the expected behavior, I think. You could, of course, always implement additional keyboard controls. Opening on click instead of focus, even when triggerOnHover is true, would work as well.

We show a separate arrow button that opens the popover. So basically there is one popover trigger for on hover and one for on click for keyboard support. Having both open via keyboard would break the behaviour because a single click selects the default thumbs up emoji instead of opening the popover.

Sqrrl:
> > > > Issue: Has probably been a problem before your changes, but the trigger always needs to be able to activate the popover via keyboard controls. I would suggest to additionally trigger the popover via the focus event when triggerOnHover is true.
> > > 
> > > 
> > > What if we want to implement keyboard support in a different way? should I add a prop that disables keyboard support?
> > 
> > 
> > What do you mean with "different way"? From an accessibility perspective focus would be the expected behavior, I think. You could, of course, always implement additional keyboard controls. Opening on click instead of focus, even when triggerOnHover is true, would work as well.
> 
> We show a separate arrow button that opens the popover. So basically there is one popover trigger for on hover and one for on click for keyboard support. Having both open via keyboard would break the behaviour because a single click selects the default thumbs up emoji instead of opening the popover.

I see. Then a prop to disable the keyboard controls should be fine.

AshleyBekemeier:
> > > > > Issue: Has probably been a problem before your changes, but the trigger always needs to be able to activate the popover via keyboard controls. I would suggest to additionally trigger the popover via the focus event when triggerOnHover is true.
> > > > 
> > > > 
> > > > What if we want to implement keyboard support in a different way? should I add a prop that disables keyboard support?
> > > 
> > > 
> > > What do you mean with "different way"? From an accessibility perspective focus would be the expected behavior, I think. You could, of course, always implement additional keyboard controls. Opening on click instead of focus, even when triggerOnHover is true, would work as well.
> > 
> > 
> > We show a separate arrow button that opens the popover. So basically there is one popover trigger for on hover and one for on click for keyboard support. Having both open via keyboard would break the behaviour because a single click selects the default thumbs up emoji instead of opening the popover.
> 
> I see. Then a prop to disable the keyboard controls should be fine.

okay, added.

Sqrrl (approved):

PR #778, opened 2024-08-12
Fix swirl-file-viewer-pdf current page number

joseabernardes:
## Summary

- [BSW-118: SwirlFileViewer shows wrong page number](https://linear.app/flip/issue/BSW-118/swirlfileviewer-shows-wrong-page-number)

To solve the problem, I changed the strategy that defines the current page.
Previously, we defined the current page as the first visible page and, when the user scrolled down to the end of the file, we defined the current page as the last one (actually, that was the bug, because we were defining the second to last page and not the last one).
The new strategy defines the current page as the first most visible page, so that when we have two visible pages, the current one is the most visible, which is the standard behaviour in other file viewers, such as Chrome.
The only edge case I've found is when we have small pages (height) and we can have two full pages visible at the end of the viewer, it will consider the current one as the first one. From what I've found, this is also what Chrome is considering, so I wouldn't consider it a problem.

If you have any question reach out!

### General

- [x] The submitted code is organized and formatted according to our [💅 Style Guide](https://swirl-storybook.flip-app.dev/?path=/docs/requirements-style-guide--docs).

- [x] The changes do not contain any breaking changes.
- [x] The changes do not introduce new components that don't belong in the library (e.g. non-reusable components, highly specialized components, components including business logic)
- [x] The component documentation is updated.
- [x] The changes meet the [🤖 testing requirements](https://swirl-storybook.flip-app.dev/?path=/docs/requirements-testing--docs).
  - [ ] All tests are 🟢.
  - [ ] No static text is used.

### DIFF
--- a/.changeset/khaki-ghosts-film.md
+++ b/.changeset/khaki-ghosts-film.md
@@ -1,0 +1,7 @@
+---
+"@getflip/swirl-components": patch
+"@getflip/swirl-components-angular": patch
+"@getflip/swirl-components-react": patch
+---
+
+Fix swirl-file-viewer-pdf current page number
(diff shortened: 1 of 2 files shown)


Sqrrl (approved):
Lgtm. I like the `firstMostVisiblePage` logic.

PR #804, opened 2024-09-12
Make SwirlAppLayout navigation collapsible

Sqrrl:
## Summary

- [Ticket](https://linear.app/flip/issue/ENG-203/make-navigation-column-in-news-collapsable)
- [Storybook](#)

### DIFF
--- a/packages/swirl-components/src/components/swirl-app-layout/swirl-app-layout.tsx
+++ b/packages/swirl-components/src/components/swirl-app-layout/swirl-app-layout.tsx
@@ -5,15 +5,26 @@ import {
   EventEmitter,
   h,
   Host,
+  Listen,
   Method,
   Prop,
   State,
   Watch,
 } from "@stencil/core";
 import classnames from "classnames";
-import { debounce, isMobileViewport, prefersReducedMotion } from "../../utils";
+import {
+  debounce,
+  getDesktopMediaQuery,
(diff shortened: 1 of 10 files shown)


packages/swirl-components/src/components/swirl-app-layout/swirl-app-layout.tsx
```diff
@@ -103,17 +129,49 @@ export class SwirlAppLayout {
...
   disconnectedCallback() {
+    this.desktopMediaQuery.removeEventListener?.(
+      "change",
+      this.desktopMediaQueryHandler
+    );
+
     this.mutationObserver?.disconnect();
   }
 
+  @Listen("click", { target: "document" })
```
AshleyBekemeier, line 153: **question (non-blocking):** does this auto-remove the event listener when the component gets destroyed?

AshleyBekemeier (approved):

packages/swirl-components/src/components/swirl-app-layout/swirl-app-layout.tsx
Sqrrl: Yes. From the Stencil docs:

> The Listen() decorator is for listening to DOM events, including the ones dispatched from @Events. The event listeners are automatically added and removed when the component gets added or removed from the DOM.

PR #849, opened 2024-11-12
getflip.dev - Fix display of nested arrays

tobias-bahls:
## Summary

Ticket: ACT-632

This is a follow-up to #844. It turned out that the rendering of properties nested in arrays was not fully implemented, causing a number of rendering problems, one of them us not being able to hide fields that are going to be deprecated.

This PR fixes this and alongside a number of other rendering issues with properties nested in arrays.


webspaceadam (approved):
LGTM :)

PR #868, opened 2024-11-20
Adjust the swirl-modal for post preview

matvlaovi-flip:
The `post-preview` requires specific styles for the secondary content configuration. Added the option to specify the widths and flex properties of both the primary and secondary content, along with removing some paddings/borders.

Fixed the issue that the `swirl-image-grid` had, as it would not remove overlays (the +2 blurred text) as items were dynamically added/removed.

Added a box shadow to `the swirl-toggle-group`

### DIFF
--- a/packages/swirl-components/src/components/swirl-image-grid/swirl-image-grid.tsx
+++ b/packages/swirl-components/src/components/swirl-image-grid/swirl-image-grid.tsx
@@ -23,6 +23,10 @@ export class SwirlImageGrid {
       this.el.children
     ) as HTMLSwirlImageGridItemElement[];
 
+    this.items.forEach((item) => {
+      item.overlay = undefined;
+    });
+
     if (this.items.length > 4) {
       this.items[3].overlay = `+${this.items.length - 4}`;
     }
--- a/packages/swirl-components/src/components/swirl-modal/swirl-modal.css
+++ b/packages/swirl-components/src/components/swirl-modal/swirl-modal.css
@@ -433,12 +433,6 @@
   }
 }
 
-.modal--remove-secondary-content-padding {
-  & .modal__secondary-content {
-    padding: 0;
-  }
-}
-
 @keyframes modal-scale-in {
   from {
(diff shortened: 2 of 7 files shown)


packages/swirl-components/src/components/swirl-image-grid/swirl-image-grid.tsx
```diff
@@ -23,6 +23,10 @@ export class SwirlImageGrid {
       this.el.children
     ) as HTMLSwirlImageGridItemElement[];
 
+    this.items.forEach((item) => {
+      item.overlay = undefined;
```
Sqrrl, line 27: Issue: This would break other usages, right?

matvlaovi-flip: What usages do you mean? 🤔 
It still sets the overlay in the following lines, but now it resets them before doing that

Sqrrl: Just saw your description, and the lines below. Nevermind.

packages/swirl-components/src/components/swirl-modal/swirl-modal.css
```diff
@@ -429,6 +433,12 @@
   }
 }
 
+.modal--remove-secondary-content-padding {
```
Sqrrl, line 436: Idea: Works for your use case. Adding a prop to control the padding, instead of just disabling it would be more flexible.

Sqrrl: Like swirl-box does for example. 
`@Prop() padding?: SwirlBoxPadding = "0";`

matvlaovi-flip: Hmm, nice true, I'll add it

matvlaovi-flip: added the options, works locally

PR #874, opened 2024-11-26
Additional `swirl-modal` options & `aria-label` in `swirl-switch`

matvlaovi-flip:
Added the option to modify the `min-height` and the `gap` of modal content containers.
Added the option to define the `aria-label` for `swirl-switch`.

### DIFF
--- a/packages/swirl-components/src/components/swirl-modal/swirl-modal.tsx
+++ b/packages/swirl-components/src/components/swirl-modal/swirl-modal.tsx
@@ -49,11 +49,13 @@ export class SwirlModal {
   @Prop() hideLabel?: boolean;
   @Prop() label!: string;
   @Prop() maxHeight?: string;
+  @Prop() minHeight?: string;
   @Prop() maxWidth?: string;
   @Prop() padded?: boolean = true;
   @Prop() primaryActionLabel?: string;
   @Prop() secondaryActionLabel?: string;
   @Prop() variant?: SwirlModalVariant = "default";
+  @Prop() contentGap?: string;
   @Prop() hideSecondaryContent?: boolean;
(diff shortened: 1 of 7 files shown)


packages/swirl-components/src/components/swirl-modal/swirl-modal.tsx
```diff
@@ -49,11 +49,13 @@ export class SwirlModal {
   @Prop() hideLabel?: boolean;
   @Prop() label!: string;
   @Prop() maxHeight?: string;
+  @Prop() minHeight?: string;
   @Prop() maxWidth?: string;
   @Prop() padded?: boolean = true;
   @Prop() primaryActionLabel?: string;
   @Prop() secondaryActionLabel?: string;
   @Prop() variant?: SwirlModalVariant = "default";
+  @Prop() contentGap?: string;
```
Sqrrl, line 58: Question: Should we limit this to token variables like we did for the swirl-stack?

```ts
export type SwirlStackSpacing =
  | "0"
  | "2"
  | "4"
  | "8"
  | "12"
  | "16"
  | "24"
  | "32"
  | "40"
  | "48"
  | "64";
  ```

Sqrrl (approved):

packages/swirl-components/src/components/swirl-modal/swirl-modal.tsx
matvlaovi-flip: Oh true, definitely.

Btw why does every component define spacings for itself, can we have a shared type with these 8-system spacings?

Sqrrl: Probably. It just varies slightly from component to component. Not every component is allowed to use all spacings.

PR #894, opened 2024-12-17
ENG-511 Image intersection issues

matvlaovi-flip:
## Summary
- [Ticket](https://linear.app/flip/issue/ENG-511/check-intersecting-loading-swirl-image-grid-inside-chat)

This solution is based on two ideas:

1. As clients (Angular in our case) render the DOM, these images might at some point be intersecting with the viewport, while it's not visible to the user. Instead of checking all of the intersection entries, we should only check the last one, as it's the "final" state that the user sees.
2. Sometimes browsers won't be able to bundle these intersection entries together, e.g. when you manually reload. This is why I also introduced a debounce so that we again only handle the last state.

Additionally sorted the entries as recommended by MDN

### DIFF
--- a/packages/swirl-components/src/components/swirl-image-grid-item/swirl-image-grid-item.tsx
+++ b/packages/swirl-components/src/components/swirl-image-grid-item/swirl-image-grid-item.tsx
@@ -11,6 +11,7 @@ import {
   Watch,
 } from "@stencil/core";
 import classnames from "classnames";
+import { debounce } from "../../utils";
 
 export type SwirlImageGridItemLoading =
   | "lazy"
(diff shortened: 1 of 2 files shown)


packages/swirl-components/src/components/swirl-image-grid-item/swirl-image-grid-item.tsx
```diff
@@ -102,19 +103,20 @@ export class SwirlImageGridItem {
...
     this.intersectionObserver.observe(this.el);
   }
 
-  private onVisibilityChange(entries: IntersectionObserverEntry[]) {
-    this.inViewport = entries.some((entry) => entry.isIntersecting);
-  }
+  private handleIntersectionEntries = debounce(
+    (entries: IntersectionObserverEntry[]) => {
+      const sorted = [...entries].sort((a, b) => a.time - b.time);
+      this.inViewport = sorted[sorted.length - 1].isIntersecting;
```
Sqrrl, line 116: Just personal preference. Ignore if you disagree. `sorted.at(-1).isIntersecting`

packages/swirl-components/src/components/swirl-image-grid-item/swirl-image-grid-item.tsx
```diff
@@ -102,19 +103,20 @@ export class SwirlImageGridItem {
...
 
     this.intersectionObserver.observe(this.el);
   }
 
-  private onVisibilityChange(entries: IntersectionObserverEntry[]) {
-    this.inViewport = entries.some((entry) => entry.isIntersecting);
-  }
+  private handleIntersectionEntries = debounce(
+    (entries: IntersectionObserverEntry[]) => {
+      const sorted = [...entries].sort((a, b) => a.time - b.time);
```
Sqrrl, line 115: Didn't consider multiple entries. Nice catch.

Sqrrl (approved):

packages/swirl-components/src/components/swirl-image-grid-item/swirl-image-grid-item.tsx
matvlaovi-flip, line 116:
```suggestion
      this.inViewport = sorted.at(-1).isIntersecting;
```

matvlaovi-flip: agreed

PR #896, opened 2024-12-17
Determine first day of week for date picker via locale

Sqrrl:
## Summary

- [Ticket](https://linear.app/flip/issue/ENG-659/scheduling-posts-starts-on-a-monday)
- [Storybook](https://icy-water-049ec4003-896.westeurope.3.azurestaticapps.net/?path=/docs/components-swirldatepicker--docs)

### DIFF
--- a/packages/swirl-components/src/components/swirl-date-picker/swirl-date-picker.tsx
+++ b/packages/swirl-components/src/components/swirl-date-picker/swirl-date-picker.tsx
@@ -49,7 +49,7 @@ export class SwirlDatePicker {
       this.firstDayOfWeek =
         "getWeekInfo" in locale
           ? locale.getWeekInfo().firstDay
-          : locale.weekInfo?.firstDay;
+          : locale.weekInfo?.firstDay ?? 0;
     }
   }
 
--- a/packages/swirl-components/src/components/swirl-date-picker/swirl-date-picker.tsx
+++ b/packages/swirl-components/src/components/swirl-date-picker/swirl-date-picker.tsx
@@ -9,9 +9,20 @@ import {
 } from "@stencil/core";
 import { WcDatepickerCustomEvent } from "wc-datepicker/dist/types/components";
 import { WCDatepickerLabels } from "wc-datepicker/dist/types/components/wc-datepicker/wc-datepicker";
+import { getISODateString, removeTimezoneOffset } from "../../utils";
 
 import "wc-datepicker";
-import { getISODateString, removeTimezoneOffset } from "../../utils";
+
+interface WeekInfo {
+  firstDay: number;
+  weekendStart: number;
+  weekendEnd: number;
+}
+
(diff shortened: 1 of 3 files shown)


packages/swirl-components/src/components/swirl-date-picker/swirl-date-picker.tsx
```diff
@@ -9,9 +9,20 @@ import {
...
+import { getISODateString, removeTimezoneOffset } from "../../utils";
 
 import "wc-datepicker";
-import { getISODateString, removeTimezoneOffset } from "../../utils";
+
+interface WeekInfo {
+  firstDay: number;
+  weekendStart: number;
+  weekendEnd: number;
+}
```
matvlaovi-flip, line 20: Don't think the object will look like that:
image

Maybe just keep the `firstDay` in it and add a comment above for the source? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo

packages/swirl-components/src/components/swirl-date-picker/swirl-date-picker.tsx
```diff
@@ -31,6 +42,17 @@ export class SwirlDatePicker {
...
   @Event({ bubbles: false }) valueChange: EventEmitter<Date | Date[]>;
 
+  componentWillLoad() {
+    if (!this.firstDayOfWeek) {
+      const locale = new Intl.Locale(this.locale) as LocaleWithWeekInfo;
+
+      this.firstDayOfWeek =
+        "getWeekInfo" in locale
+          ? locale.getWeekInfo().firstDay
+          : locale.weekInfo?.firstDay;
```
matvlaovi-flip, line 52: Doesn't this mean that for firefox firstDayOfWeek will be undefined as none of those properties exist? Maybe add a final default which is 0?

packages/swirl-components/src/components/swirl-date-picker/swirl-date-picker.tsx
Sqrrl: Haha, the one time I trust ChatGPT. 🫣 Should have known better. Done.

packages/swirl-components/src/components/swirl-date-picker/swirl-date-picker.tsx
Sqrrl: Tested it and should be fine. It defaults to 0 in https://sqrrl.github.io/wc-datepicker/

Sqrrl: It would be passed to `<wc-datepicker />` as  `undefined`. So it would take its default. https://github.com/Sqrrl/wc-datepicker/blob/main/src/components/wc-datepicker/wc-datepicker.tsx

Added the fallback anyway to be safe.

matvlaovi-flip (approved):

PR #930, opened 2025-02-04
ENG-877 Add pill variant to `swirl-tab-bar`

matvlaovi-flip:
## Summary

`swirl-tab-bar` now supports a pill variant, additional `paddingBlockStart` and `paddingBlockEnd` properties, and can be horizontally scrolled instead of a previously faulty truncation logic.

- [Ticket](https://linear.app/flip/issue/ENG-877/update-tabs-on-web)
- [Design](https://www.figma.com/design/iRHiAFsHxnzux8gQ9EjlZe/Web-Components?node-id=34-6419&p=f&m=dev)
- [Storybook](https://icy-water-049ec4003-930.westeurope.3.azurestaticapps.net/?path=/docs/components-swirltabbar--docs)


packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.tsx
```diff
@@ -30,9 +32,13 @@ export class SwirlTabBar {
   @Prop() disableTabSemantics?: boolean;
   @Prop() label!: string;
   @Prop() justify?: SwirlTabBarJustify = "start";
+  @Prop() paddingBlockEnd?: SwirlTabBarPadding;
+  @Prop() paddingBlockStart?: SwirlTabBarPadding;
```
Sqrrl, line 36: **question:** Should we enforce a padding for the new variant to not clip the focus indicator?

matvlaovi-flip: Good question, in all the designs we have these paddings set around so clipping won't happen, but if we enforce it it would be a breaking change as technically the `padding-inline-start` would also be necessary. 
image

matvlaovi-flip: image

our left menu has the same problem

matvlaovi-flip: Maybe we could leave a warning for it somewhere in the docs?

matvlaovi-flip: Interesting discussion here https://github.com/w3c/csswg-drafts/issues/3417

It seems the conclusion is: adjust your layout with some paddings to take care of it. It feels weird for us to enforce it here tho

Sqrrl: Makes sense.

Sqrrl (approved):

PR #936, opened 2025-02-05
Drag & Drop support for swirl-tree-view

Sqrrl:
## Summary

- [Ticket](https://linear.app/flip/issue/ENG-872/spike-pages-drag-and-drop)
- [Storybook](https://icy-water-049ec4003-936.westeurope.3.azurestaticapps.net/?path=/story/components-swirltreeview--swirl-tree-view)

### DIFF
--- a/packages/swirl-components/src/components/swirl-tree-view/swirl-tree-view.config.ts
+++ b/packages/swirl-components/src/components/swirl-tree-view/swirl-tree-view.config.ts
@@ -4,6 +4,7 @@ export const treeViewDragDropConfig: Partial<Sortable.Options> = {
   animation: 100,
   chosenClass: "tree-view-item--chosen",
   direction: "vertical",
+  dragClass: "tree-view-item--drag",
   draggable: "swirl-tree-view-item",
   emptyInsertThreshold: 0,
   fallbackOnBody: true,
--- a/packages/swirl-components/src/components/swirl-tree-view/swirl-tree-view.config.ts
+++ b/packages/swirl-components/src/components/swirl-tree-view/swirl-tree-view.config.ts
@@ -18,5 +18,6 @@ export const treeViewDragDropConfig: Partial<Sortable.Options> = {
   onStart: (event) => {
     event.item.closest(".tree-view")?.classList.add("tree-view--dragging");
   },
-  swapThreshold: 0.5,
+  swapThreshold: 0.75,
+  invertedSwapThreshold: 1,
 };
--- a/packages/swirl-components/src/components/swirl-tree-view/swirl-tree-view.config.ts
+++ b/packages/swirl-components/src/components/swirl-tree-view/swirl-tree-view.config.ts
@@ -1,0 +1,16 @@
+import type Sortable from "sortablejs";
+
+export const treeViewDragDropConfig: Partial<Sortable.Options> = {
+  animation: 150,
+  chosenClass: "tree-view-item--chosen",
+  draggable: "swirl-tree-view-item",
+  fallbackOnBody: true,
+  ghostClass: "tree-view-item--ghost",
+  group: "swirl-tree-view",
+  onEnd: (event) => {
+    event.item.closest(".tree-view")?.classList.remove("tree-view--dragging");
+  },
+  onStart: (event) => {
+    event.item.closest(".tree-view")?.classList.add("tree-view--dragging");
+  },
+};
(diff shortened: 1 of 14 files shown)


packages/swirl-components/src/components/swirl-tree-view/swirl-tree-view.tsx
```diff
@@ -295,14 +432,64 @@ export class SwirlTreeView {
     return (nestedSibling as HTMLSwirlTreeViewItemElement) ?? undefined;
   }
 
+  private updateLiveRegionText(
```
Sqrrl, line 435: **note:** I couldn't find an agreed upon pattern for accessible drag&drop via keyboard. The live region instructions might be a bit verbose. Not sure.

matvlaovi-flip:
image
 
 Drag preview seems a bit broken

matvlaovi-flip:
Not sure if by design but it's a bit difficult figuring out whether the item will be dropped next to an item or as a child of it.
Also I assumed that if you drag onto an item with collapsed children, that the item will expand and you can see all the children

matvlaovi-flip (approved):
I'd let Torben and Liu check the behavior but code looks great, very impressive work here

Sqrrl:
> image
> Drag preview with multiple items seems a bit broken

Fixed.

Sqrrl:
> Not sure if by design but it's a bit difficult figuring out whether the item will be dropped next to an item or as a child of it. Also I assumed that if you drag onto an item with collapsed children, that the item will expand and you can see all the children

Yes. It's still not ideal. Already talked quite a lot about that with Liu. The "drop on top" isn't supported out of the box. I'm sure there is a way to build it, but I don't see a way without building around the Sortable.js stuff.

The problem right now with the dropping inside a leaf node is the drop zone of that leaf. Sortable.js needs the children container (drop zone) to be in the DOM and also to occupy "physical" space (doesn't work with zero height or absolute positioned containers). That's why I have that awkward 0.5rem height container with negative 0.5rem margin for leaf nodes. Making that container larger or moving it up and down messes with other drop zones too much. I also fiddled a lot with the `swapThreshold` value to improve the behaviour, but couldn't improve it any further. Maybe you can give it a try and find a nicer solution.

packages/swirl-components/src/components/swirl-tree-view/swirl-tree-view.config.ts
```diff
@@ -0,0 +1,22 @@
...
+  group: "swirl-tree-view",
+  invertSwap: true,
+  onEnd: (event) => {
+    event.item.closest(".tree-view")?.classList.remove("tree-view--dragging");
+  },
+  onStart: (event) => {
+    event.item.closest(".tree-view")?.classList.add("tree-view--dragging");
+  },
+  swapThreshold: 0.75,
+  invertedSwapThreshold: 1,
```
AshleyBekemeier, line 21: **issue:** The `swapThreshold` property is ignored because the `invertedSwapThreshold` and `invertSwap` properties are both set. Delete one of them and consider using a lower threshold to make the tree feel a bit more _rigid_

Sqrrl: I thought so too and maybe I'm imagining stuff, but changing the swapThreshold still has an effect, event if `invertSwap` is `true`. The current values were the best compromise I could come up with to make the "drop into leaf node" not too difficult. What values would you suggest?

packages/swirl-components/src/components/swirl-tree-view/swirl-tree-view.tsx
```diff
@@ -46,6 +93,38 @@ export class SwirlTreeView {
...
+      `) as HTMLSwirlTreeViewItemElement | undefined;
+
+      if (oldParentItem) {
+        forceUpdate(oldParentItem);
+        oldParentItem.expand();
+      }
+    }
+  }
+
   @Listen("keydown")
```
AshleyBekemeier, line 128: **nitpick:** can we prevent the screen from scrolling down when pressing space to move items?

Sqrrl: Yes. Done. Good catch.

packages/swirl-components/src/components/swirl-tree-view/swirl-tree-view.config.ts
AshleyBekemeier: from my testing it doesn't seem to affect the drag and drop? Maybe i am missing a case but i couldn't tell a difference with any value set to it. Unless I am mistaken it can be deleted then.

And i would set the `invertedSwapThreshold` to something lower than 1, maybe like `0.5`, to allow the user more space to drag items as children

Sqrrl: invertedSwapThreshold 0.75 seems to work. 0.5 made some swaps impossible.

PR #937, opened 2025-02-05
Adjust context menus

AshleyBekemeier:
## Summary

- [Ticket](https://linear.app/flip/issue/ENG-829/adjust-context-menus-swirl-components)


packages/swirl-components/src/components/swirl-action-list/swirl-action-list.tsx
```diff
@@ -24,6 +24,7 @@ export class SwirlActionList {
         this.container.setAttribute("tabindex", "0");
       }
     });
+    this.setSectionSeparator();
```
matvlaovi-flip, line 27: Do you think we should also call this on the `slotchange` event, as depending on how this is used items could be changed after the component was loaded. Same question for the option list

matvlaovi-flip (approved):

packages/swirl-components/src/components/swirl-action-list/swirl-action-list.tsx
AshleyBekemeier: added

PR #964, opened 2025-03-04
ENG-1110 Add `none` option to `swirl-skeleton-box` border-radius

matvlaovi-flip:


packages/swirl-components/src/components/swirl-skeleton-box/swirl-skeleton-box.tsx
```diff
@@ -1,7 +1,12 @@
 import { Component, h, Host, Prop } from "@stencil/core";
 import classnames from "classnames";
 
-export type SwirlSkeletonBoxBorderRadius = "s" | "sm" | "base" | "pill";
+export type SwirlSkeletonBoxBorderRadius =
+  | "s"
+  | "sm"
+  | "base"
+  | "pill"
+  | "null";
```
AshleyBekemeier, line 9: **suggestion:** this is a string of the word `null`? Why not add actually make it nullable. (also what about undefined?)

matvlaovi-flip: Did not want to interfere with the current default value `base`.
Passing `undefined` to the prop also didn't feel right. In this case it might look weird but it's the actual name of the token for border-radius

Sqrrl: I think we have similar cases where we used "none" instead of "null".

matvlaovi-flip: image

This is how it's set, using the token. For `pill` there's a class that handles it. Let me know if you want me to adjust it

AshleyBekemeier: i'm not saying it doesn't work nicely, it's just weird from the perspective of someone using this component. Putting a string of `null` in there seems like it would be a mistake. I also then prefer what philipp suggested with 'none'

matvlaovi-flip: I understand, just didn't like the raised complexity, pls check now

Sqrrl (approved):

PR #965, opened 2025-03-04
Drag & drop for swirl-table

Sqrrl:
https://icy-water-049ec4003-965.westeurope.3.azurestaticapps.net/?path=/story/components-swirltable--swirl-table

### DIFF
--- a/packages/swirl-components/src/components.d.ts
+++ b/packages/swirl-components/src/components.d.ts
@@ -2214,6 +2214,7 @@ export namespace Components {
     interface SwirlTable {
         "caption"?: string;
         "dragDropHandle"?: string;
+        "dragDropInstructions": { end: string; initial: string; moved: string; start: string; };
         "emptyStateLabel"?: string;
         "enableDragDrop"?: boolean;
         "label": string;
@@ -7897,6 +7898,7 @@ declare namespace LocalJSX {
     interface SwirlTable {
         "caption"?: string;
         "dragDropHandle"?: string;
+        "dragDropInstructions"?: { end: string; initial: string; moved: string; start: string; };
         "emptyStateLabel"?: string;
         "enableDragDrop"?: boolean;
         "label": string;
--- a/packages/swirl-components/src/components.d.ts
+++ b/packages/swirl-components/src/components.d.ts
@@ -63,6 +63,7 @@ import { SwirlSwitchLabelPosition } from "./components/swirl-switch/swirl-switch
 import { SwirlSymbolSize } from "./components/swirl-symbol/swirl-symbol.types";
 import { SwirlTabPadding } from "./components/swirl-tab/swirl-tab";
 import { SwirlTabBarJustify, SwirlTabBarPadding, SwirlTabBarTab, SwirlTabBarVariant } from "./components/swirl-tab-bar/swirl-tab-bar";
+import { SwirlTableDropRowEvent } from "./components/swirl-table/swirl-table";
 import { SwirlTableColumnSort } from "./components/swirl-table-column/swirl-table-column";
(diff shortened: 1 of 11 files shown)


packages/swirl-components/src/components/swirl-table/swirl-table.tsx
```diff
@@ -3,46 +3,85 @@ import debouncePromise from "debounce-promise";
...
+    "Row grabbed. Current position in table: {position} of {rowCount}. Press up and down arrow keys to change position, Space to drop.",
+};
 
 /**
  * @slot columns - Column container, should contain SwirlTableColumns.
  * @slot rows - Row container, should contain SwirlTableRows.
  */
 @Component({
-  shadow: true,
+  shadow: false,
```
matvlaovi-flip, line 42: Probably not in our case but generally this would be a breaking change right? If some clients reach parts of the table within the shadow dom

packages/swirl-components/src/components.d.ts
```diff
@@ -2211,7 +2213,10 @@ export namespace Components {
     }
     interface SwirlTable {
         "caption"?: string;
+        "dragDropHandle"?: string;
```
AshleyBekemeier, line 2216: **nitpick:** maybe a short description would be nice for this one

packages/swirl-components/src/components/swirl-table/swirl-table.tsx
```diff
@@ -396,14 +591,68 @@ export class SwirlTable {
...
+    if (this.el.contains(newlyFocusedElement) && focusedDragHandle) {
+      return;
+    }
+
+    if (this.liveRegionText !== "") {
+      this.updateLiveRegionText();
+    }
+  };
+
+  private onKeyDown = (event: KeyboardEvent) => {
```
matvlaovi-flip, line 619: Shouldn't we check if the drag handle was focused before going into drag mode?
Focusing any cell and pressing space starts the drag thing and then focus moves to the handle after dragging. I would expect to be able to enter the right menu of the row, or to select checkbox via keyboard?

packages/swirl-components/src/components/swirl-table/swirl-table.tsx
```diff
@@ -61,6 +100,61 @@ export class SwirlTable {
...
+      }
+
+      if (!this.dragDropHandle) {
+        console.warn(
+          '[Swirl] swirl-table required the "dragDropHandle" prop to be set when drag & drop is enabled.'
+        );
+        return;
+      }
+
+      this.sortable = new Sortable(this.bodyEl, {
```
matvlaovi-flip, line 129: table drop animations

Does sortable handle the animation? Looks like it's going to the wrong spot in some cases as seen in the gif

packages/swirl-components/src/components/swirl-table/swirl-table.tsx
Sqrrl: Yes. Risk is minimal I think, because we are only using the table in the AC. I don't know about any external usage.

packages/swirl-components/src/components/swirl-table/swirl-table.tsx
Sqrrl: Yes. Done.

packages/swirl-components/src/components/swirl-table/swirl-table.tsx
Sqrrl: This is sortable. Probably has to do with the "spill" beahviour.

matvlaovi-flip:
image

If you enter the drag mode with `space`, and then unfocus the table by clicking somewhere outside with a mouse, the focused style remains on the row

matvlaovi-flip:
When we use the keyboard to drag the rows up,down we also scroll the entire window, does it make sense to fix that?

matvlaovi-flip:
image

Preview of the row looks a bit funky, probably because of the fixed columns?

matvlaovi-flip (approved):
Other than some minor visual stuff works beautifully 👏

PR #970, opened 2025-03-06
Replace onLoad/onError img attributes with addEventListener subscriptions

Sqrrl:
## Summary

- [Ticket](https://linear.app/flip/issue/ENG-1151/fix-swirl-onloadonerror-listeners)
- [Storybook](#)

### DIFF
--- a/packages/swirl-components/src/components/swirl-app-icon/swirl-app-icon.tsx
+++ b/packages/swirl-components/src/components/swirl-app-icon/swirl-app-icon.tsx
@@ -12,12 +12,31 @@ export class SwirlAppIcon {
 
   @State() imageAvailable: boolean | undefined;
 
+  private imgEl: HTMLImageElement | undefined;
+
+  disconnectedCallback() {
+    this.imgEl?.removeEventListener("load", this.setImageAvailable);
+    this.imgEl?.removeEventListener("error", this.setImageUnavailable);
+  }
+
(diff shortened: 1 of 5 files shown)


packages/swirl-components/src/components/swirl-app-icon/swirl-app-icon.tsx
```diff
@@ -12,6 +12,13 @@ export class SwirlAppIcon {
 
   @State() imageAvailable: boolean | undefined;
 
+  private imgEl: HTMLImageElement | undefined;
+
+  disconnectedCallback() {
```
matvlaovi-flip, line 17: are you sure we need the `disconnectedCallback()`, doesn't the `ref` also trigger the callback on unmount, and you already remove the event listeners there?

Sqrrl: Nope. Checked like this in the story:

```
setTimeout(() => {
    element.parentElement.removeChild(element);
  }, 3000);
```

The ref callback was not called.

Sqrrl: Guess it makes sense, looking at the diagram here: https://stenciljs.com/docs/component-lifecycle

matvlaovi-flip (approved):
Let's see if it gets rid of the errors

PR #972, opened 2025-03-07
Fix detached box issue

Sqrrl:


packages/swirl-components/src/components/swirl-box/swirl-box.tsx
```diff
@@ -102,6 +102,10 @@ export class SwirlBox {
   };
 
   private updateScrollState() {
+    if (!this.el.isConnected) {
```
Sqrrl, line 105: **note:** Lines 95 to 97 can cause this code to be run after the component was removed from the DOM, causing the exceptions. I wasn't able to find a better trigger for the initial updateScrollState call, so I added this check here.

matvlaovi-flip (approved):

PR #1054, opened 2025-05-08
Hide closed SwirlPopovers entirely

Sqrrl:
## Summary

- [Ticket](https://linear.app/flip/issue/ENG-1691/save-button-moves-out-of-view-port-when-more-and-more-images-are)
- [Storybook](#)

The popover always had a `display: inline-flex` which could mess up the layout flow of pages, even when the popover was closed. This change applies `display: none` to closed popovers.

### DIFF
--- a/.changeset/poor-walls-complain.md
+++ b/.changeset/poor-walls-complain.md
@@ -1,0 +1,8 @@
+---
+"@getflip/swirl-components": patch
+"@getflip/swirl-components-angular": patch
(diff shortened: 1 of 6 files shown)


Sqrrl:
@joseabernardes Github actions are timing out right now. Not sure why, but maybe you can still review and check the changes locally.

joseabernardes (approved):

PR #1056, opened 2025-05-12
Fix yarn install in Github actions

Sqrrl:
One of the post-install scripts (of a dependency) is timing out in Github actions, so I …

- added `yarn install --ignore-scripts` to the workflows
- moved the `patch-package` script from the `post-install` to the `prebuild` step

In my manual tests omitting the post-install scripts (other than `patch-package`) of the dependencies seemed fine. For the `publish` workflow we need to merge and check on main.

A nice side-effect: This speeds up the workflows by a lot.

### DIFF
--- a/.github/workflows/build-deploy.yml
+++ b/.github/workflows/build-deploy.yml
@@ -30,11 +30,12 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: true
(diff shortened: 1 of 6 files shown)


joseabernardes (approved):
Let's see if we don't missing anything important from the post-install of the dependencies 🤞

PR #1059, opened 2025-05-13
add aria-live to character counter

AshleyBekemeier:
## Summary

- [Ticket](https://linear.app/flip/issue/EMP-110/11413-dynamically-updating-content-not-announced-for-screen-reader)

### DIFF
--- a/packages/swirl-components/src/components/swirl-text-input/swirl-text-input.tsx
+++ b/packages/swirl-components/src/components/swirl-text-input/swirl-text-input.tsx
@@ -389,7 +389,7 @@ export class SwirlTextInput implements SwirlFormInput {
             </span>
           )}
           {this.showCharacterCounter && (
-            <span class="text-input__character-counter">
+            <span class="text-input__character-counter" aria-live="polite">
               {this.value?.length || 0}{" "}
               {Boolean(this.maxLength) ? `/ ${this.maxLength}` : ""}
             </span>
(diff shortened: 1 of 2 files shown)


matvlaovi-flip (approved):

packages/swirl-components/src/components/swirl-text-input/swirl-text-input.tsx
```diff
@@ -389,7 +389,7 @@ export class SwirlTextInput implements SwirlFormInput {
             </span>
           )}
           {this.showCharacterCounter && (
-            <span class="text-input__character-counter">
+            <span class="text-input__character-counter" aria-live="polite">
```
Sqrrl, line 392: **question:** Not sure about this, but it might be necessary to also add a (hidden) label here. E.g. 

```js
<span class="text-input__character-counter" aria-live="polite">
  Used characters:
  {this.value?.length || 0}{" "}
  {Boolean(this.maxLength) ? `/ ${this.maxLength}` : ""}
</span>
```

With a translatable prop of course.

matvlaovi-flip: Also the 5 / 7 might be read as "five slash seven" so maybe we should aria-hide the content of this, and have the visually hidden part contain a sentence e.g. "5 out 7 characters used"

Sqrrl: I guess in this case the aria-live element should also be part of the `aria-describedby` attribute of the input element.

AshleyBekemeier: You're right, there's context missing. added a characterCounterLabel.

PR #1079, opened 2025-05-28
EMP-257 Improve media query listeners

matvlaovi-flip:
## Summary

- [Ticket](https://linear.app/flip/issue/EMP-257/swirl-refactor-desktopmediaquery-listeners)

We were registering way too many media query listeners; when I did some tests the number was around ~550 of them when on the newsfeed on `master`.
This approach only registers one native event listener, and our logic is in a single place that gives us full control.
Additionally, whenever you subscribe to the service you immediately get the current value, so components only need to subscribe to it and respond to the changes.

Additionally, icons used to flicker as described here https://linear.app/flip/issue/EMPWEB-12/tasks-icons-flickers-when-you-navigate-to-the-page
This doesn't seem to be the case anymore after this.

### DIFF
--- a/packages/swirl-components/src/services/media-query.service.ts
+++ b/packages/swirl-components/src/services/media-query.service.ts
@@ -1,0 +1,28 @@
+import { getDesktopMediaQuery } from "../utils";
+
+type CallbackFn = (isDesktop: boolean) => void;
+
+class DesktopMediaQueryService {
+  private mql: MediaQueryList;
+  private listeners = new Set<CallbackFn>();
+
+  constructor() {
+    this.mql = getDesktopMediaQuery();
+    this.mql.addEventListener("change", this.onChange);
+  }
+
+  private onChange = (e: MediaQueryListEvent) => {
+    this.listeners.forEach((callback) => callback(e.matches));
+  };
+
+  public subscribe(callback: CallbackFn): () => void {
+    this.listeners.add(callback);
+    callback(this.mql.matches);
+
+    return () => {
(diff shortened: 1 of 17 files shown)


packages/swirl-components/src/services/media-query.service.ts
```diff
@@ -0,0 +1,28 @@
+import { getDesktopMediaQuery } from "../utils";
+
+type CallbackFn = (isDesktop: boolean) => void;
+
+class DesktopMediaQueryService {
+  private mql: MediaQueryList;
```
AshleyBekemeier, line 6: **nitpick:** I generally don't like abbreviations for variables

packages/swirl-components/src/services/media-query.service.ts
```diff
@@ -0,0 +1,28 @@
...
+  constructor() {
+    this.mql = getDesktopMediaQuery();
+    this.mql.addEventListener("change", this.onChange);
+  }
+
+  private onChange = (e: MediaQueryListEvent) => {
+    this.listeners.forEach((callback) => callback(e.matches));
+  };
+
+  public subscribe(callback: CallbackFn): () => void {
```
AshleyBekemeier, line 18: **nitpick:** The function is called subscribe, which makes it look like it's a rxjs subscription when it's not. Let's try to avoid defined keywords as names

packages/swirl-components/src/services/media-query.service.ts
matvlaovi-flip: Sure 👍

PR #1140, opened 2025-07-24
feat(components): make SwirlPopover a valid value for SwirlPopoverTri…

Sqrrl:
In Angular it is easier to pass the SwirlPopover component instance to the swirl-popover-trigger. With these changes instead of this …

```
<swirl-popover #popover></swirl-popover>
<swirl-popover-trigger [swirlPopover]="$any(popover).el"></swirl-popover-trigger>
```

… we can do this …

```
<swirl-popover #popover></swirl-popover>
<swirl-popover-trigger [swirlPopover]="popover"></swirl-popover-trigger>
```

And not get errors like `popoverEl.addEventListener is not a function`.

### DIFF
--- a/.changeset/metal-balloons-refuse.md
+++ b/.changeset/metal-balloons-refuse.md
@@ -1,0 +1,8 @@
+---
+"@getflip/swirl-components": minor
+"@getflip/swirl-components-angular": minor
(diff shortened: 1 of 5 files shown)


joseabernardes (approved):
Lgtm :)

PR #1191, opened 2025-08-22
feat(components): add support for 3+ avatars in avatar group

Sqrrl:
## Summary

- [Ticket](https://linear.app/flip/issue/ENG-2334/flipweb-add-page-seen-button)
- [Design](https://www.figma.com/design/t6ISOeXcBoEErcolnLivai/Knowledge-Base?node-id=2491-35895&m=dev)

### DIFF
--- a/.changeset/dirty-cameras-grow.md
+++ b/.changeset/dirty-cameras-grow.md
@@ -1,0 +1,7 @@
+---
+"@getflip/swirl-components": minor
+"@getflip/swirl-components-angular": minor
+"@getflip/swirl-components-react": minor
+---
+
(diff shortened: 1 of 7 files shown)


joseabernardes:
**note:** As a component this change makes sense, but for the use case that we're building this, it doesn't quite fit, because we want to be able to have the horizontal stack styles also for 1 or 2 avatars.

Sqrrl:
> **note:** As a component this change makes sense, but for the use case that we're building this, it doesn't quite fit, because we want to be able to have the horizontal stack styles also for 1 or 2 avatars.

🤦‍♂️ Right. I will make it an explicit prop.

PR #1222, opened 2025-09-10
Adjust resource list item hover behavior

joseabernardes:
## Summary

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### General

  - [ ] All tests are 🟢.
  - [ ] No static text is used.

### DIFF
--- a/packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.css
+++ b/packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.css
@@ -218,10 +218,7 @@
   width: 100%;
   min-width: 0;
   margin: 0;
-  padding-top: var(--s-space-12);
-  padding-right: 0;
-  padding-bottom: var(--s-space-12);
-  padding-left: var(--s-space-8);
+  padding: var(--s-space-12) var(--s-space-8);
   align-items: flex-start;
   border: none;
   border-radius: var(--s-border-radius-sm);
@@ -260,7 +257,6 @@
   }
 
   @media (--from-desktop-without-touch) {
-    padding: var(--s-space-12) var(--s-space-8);
     line-height: var(--s-line-height-base);
   }
 }
--- a/packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.css
+++ b/packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.css
@@ -358,6 +358,10 @@
   flex-direction: column;
   align-self: flex-start;
   gap: var(--s-space-2);
+
+  &:not(:has(.resource-list-item__meta-text)) {
+    align-self: center;
+  }
 }
 
 .resource-list-item__meta-text {
--- a/packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.css
+++ b/packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.css
@@ -142,8 +142,11 @@
       opacity: 1;
     }
 
-    & .resource-list-item__meta {
-      pointer-events: none;
+    & .resource-list-item__badges {
+      padding-right: calc(var(--s-resource-list-item-control-size) + var(--s-space-8));
+    }
+
+    & .resource-list-item__meta-text {
       opacity: 0;
     }
   }
@@ -313,6 +316,7 @@
   right: var(--s-space-8);
   transform: translateY(-50%);
   z-index: 2;
+  transition: opacity 0.1s;
 }
 
 .resource-list-item__checkbox {
@@ -363,6 +367,7 @@
 .resource-list-item__badges {
   display: inline-flex;
   width: max-content;
+  transition: padding 0.1s;
 }
 
 .resource-list-item__drag-handle {
--- a/packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
+++ b/packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
@@ -66,6 +66,8 @@ export class SwirlResourceListItem {
   @Event() valueChange: EventEmitter<boolean>;
 
   private controlContainer: HTMLElement;
+  private badgesContainer: HTMLElement;
+  private labelContainer: HTMLElement;
   private elementId = uuid();
   private iconEl: HTMLElement;
   private mediaQueryUnsubscribe: () => void = () => {};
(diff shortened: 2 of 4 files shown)


packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
```diff
@@ -200,12 +229,6 @@ export class SwirlResourceListItem {
     const containerRole = hostRole === "row" ? "gridcell" : undefined;
 
     const labelContainerStyles = {
-      paddingRight:
-        !showMeta && Boolean(this.controlContainer)
-          ? `calc(${
-              this.controlContainer?.getBoundingClientRect().width
```
joseabernardes, line 206: **note:** Since I also needed to use the width of the control container, I've noticed that when only one render of the component is performed (storybook for example), the `controlContainer` was not defined yet, so the width couldn't be retrieved. That's why I moved this to the `componentDidRender` hook.

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
```diff
@@ -89,16 +91,43 @@ export class SwirlResourceListItem {
...
 
+  private forceStyles() {
+    const controlWidth = this.controlContainer?.getBoundingClientRect().width;
+
+    this.labelContainer?.style.setProperty(
+      "paddingRight",
+      !this.showMeta && Boolean(this.controlContainer)
+        ? `calc(${controlWidth}px + var(--s-space-16))`
+        : undefined
+    );
```
Sqrrl, line 124: **issue:** This doesn't seem to work. Did you mean?

```
this.labelContainer.style.paddingRight =
      !this.showMeta && Boolean(this.controlContainer)
        ? `calc(${controlWidth}px + var(--s-space-16))`
        : undefined;
```

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
```diff
@@ -89,16 +91,43 @@ export class SwirlResourceListItem {
...
   private forceIconProps(smallIcon: boolean) {
     const icon = this.iconEl?.children[0];
 
     icon?.setAttribute("size", smallIcon ? "20" : "24");
   }
 
+  private forceStyles() {
+    const controlWidth = this.controlContainer?.getBoundingClientRect().width;
+
+    this.labelContainer?.style.setProperty(
```
Sqrrl, line 119: **question:** We could really simplify all of this if the meta text could also stay visible on hover/when the controls are shown. Then we wouldn't have to position the control absolutely and could just render it in the same container as the meta text and bagdes. Is that an option?

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
```diff
@@ -89,16 +91,43 @@ export class SwirlResourceListItem {
     }
   }
 
+  componentDidRender() {
+    this.forceStyles();
```
Sqrrl, line 95: **issue:** Styles are not updated when the control slot changes.

E.g. in your story:

```ts
setTimeout(() => {
  element.querySelector('[slot="control"]').remove();
}, 1000);
```

or 

```ts
setTimeout(() => {
  const control = element.querySelector<HTMLElement>('[slot="control"]');

  control.style.display = "block";
  control.style.width = "100px";
}, 1000);
```

`onSlotchange` cannot be used here, because of the `shadow: false; scoped: true` on the component. So we would need an observer.

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
Sqrrl: Sorry, bit confused.

>  I decided against it […] I went for it

So you decided to keep it as is? With the width calculation and dynamic padding?

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
joseabernardes: I wanted to use `setProperty` to be able to use optional chaining in the `labelContainer`, but I should have used `padding-right` instead of `paddingRight` 🤦 

Not needed anymore.

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
joseabernardes: Nice catch :)
Non relevant anymore 🤞

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
joseabernardes: Sorry, I was not clear enough :(

I wanted to state that **initially** I decided against it. After your comment, I went for it 😄

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
```diff
@@ -180,7 +187,7 @@ export class SwirlResourceListItem {
 
     const disabled = this.disabled && !Boolean(this.href);
 
-    const hasBadges = Boolean(this.el.querySelector("[slot='badges']"));
+    const hasBadges = this.hasBadges;
```
Sqrrl, line 190: **nitpick:** Could just use the getter directly.

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
```diff
@@ -189,7 +196,7 @@ export class SwirlResourceListItem {
     const showControlOnFocus = hasControl && (Boolean(this.meta) || hasBadges);
     const showMenu =
       Boolean(this.menuTriggerId) && !Boolean(this.meta) && !this.selectable;
-    const showMeta = (Boolean(this.meta) || hasBadges) && !this.selectable;
+    const showMeta = this.showMeta;
```
Sqrrl, line 199: **nitpick:** Could just use the getter directly.

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.css
```diff
@@ -140,17 +140,23 @@
     & .resource-list-item__control {
       pointer-events: auto;
       opacity: 1;
+      width: auto;
     }
 
-    & .resource-list-item__meta {
-      pointer-events: none;
+    & .resource-list-item__badges {
+      padding-right: calc(var(--s-resource-list-item-control-size) + var(--s-space-8));
```
Sqrrl, line 147: **suggestion:** The var is no longer needed, right?

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
```diff
@@ -279,12 +280,20 @@ export class SwirlResourceListItem {
             </span>
             {showMeta && (
               <span class="resource-list-item__meta">
-                <span class="resource-list-item__meta-text">{this.meta}</span>
+                {this.meta && (
+                  <span class="resource-list-item__meta-text">{this.meta}</span>
```
Sqrrl, line 284: **suggestion:** Not sure this works with all use cases, but we could also think about always showing the meta text.
Bildschirmfoto 2025-09-10 um 13 32 09

Sqrrl (approved):
Way cleaner this way. 

Totally optional: I guess you could try to use view transitions for the animation. Never tried it for enter/leave animations, but could be an experiment.

https://developer.mozilla.org/en-US/docs/Web/CSS/::view-transition-new
https://caniuse.com/mdn-css_selectors_view-transition-new

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.css
joseabernardes: Forgot to do an overall review after my last change 😞

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
joseabernardes: Same as above

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
joseabernardes: Done :)

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
joseabernardes: Didn't do it also for nitpick reasons, to avoid running the logic in the getter multiple times 😄

PR #1606, opened 2026-04-09
Create swirl-mcp package

joseabernardes:
## Summary

Extracts the MCP server out of `@getflip/swirl-ai` into its own dedicated package, `@getflip/swirl-mcp`. The new package supports two transports (stdio and HTTP) and fetches component metadata at runtime from the `@getflip/swirl-ai` CDN artifacts, keeping the two packages cleanly separated by concern.

### `packages/swirl-mcp` (new package)
- **Two transports:** `stdio` (default, for local MCP clients) and `http` (stateless Streamable HTTP on port 3000, for remote/hosted usage)
- - **Runtime data loading:** component metadata is fetched from `unpkg.com/@getflip/swirl-ai` at request time, keyed by the user's installed `@getflip/swirl-components` version
- - **In-memory LRU cache** (`LibraryCache`) avoids redundant CDN fetches across requests
- - **6 tools:** `list_components`, `list_icons`, `list_symbols`, `list_emojis`, `get_component_details`, `get_started`
- - **Dockerfile** for deploying the HTTP transport as a container
### `packages/swirl-ai`
- MCP server code removed (scripts, tools, data loader, bin entry, tsup config)
- - README updated to point to `@getflip/swirl-mcp` for MCP usage; retains documentation for build-time artifacts


packages/swirl-mcp/src/tools/get-component-details.ts
```diff
@@ -1,23 +1,33 @@
...
+  loadLibrary: LoadLibrary
 ) {
   server.registerTool(
     "get_component_details",
     {
       description:
         "Get full details for a Swirl component including all props with types and defaults, " +
-        "events, methods, slots, accessibility info, and usage examples.",
+        "events, methods, slots, accessibility info, and usage examples. " +
+        "IMPORTANT: First read the user's package.json to determine their installed @getflip/swirl-components version, then pass it as the 'version' parameter.",
```
Sqrrl, line 21: **nitpick:** Why not a template literal for multiline string?

Sqrrl (commented):
Looks good.

Image

The remote MCP is still WIP, right?

Sqrrl (approved):

joseabernardes:
> Looks good.
> 
> Image
> The remote MCP is still WIP, right?

Partially, yes. 
The server itself is already working and can be tested locally. What we are still missing is the deployment part. I'm aligning with Platform on the approach to follow there.

packages/swirl-mcp/src/tools/get-component-details.ts
joseabernardes: Because we want to output it in a single line, but keep the code on multiple lines to not have a huge line length.
(Am I missing another way to do it? 🤔)

Sqrrl: Didn't consider that. 😅

PR #1608, opened 2026-04-13
Add trailing icon to swirl-chip

matvlaovi-flip:
## Summary

- [Ticket](https://linear.app/flip/issue/EMPWEB-354/add-right-icon-prop-to-swirl-chip)


packages/swirl-components/src/components/swirl-chip/swirl-chip.tsx
```diff
@@ -35,6 +35,7 @@ export class SwirlChip {
   @Prop() borderRadius?: SwirlChipBorderRadius = "pill";
   @Prop() icon?: string;
   @Prop() iconColor?: SwirlChipIconColor = "default";
+  @Prop() trailingIcon?: string;
```
AshleyBekemeier, line 38: It's kind of weird that the iconColor prop doesn't change the trailing icon

matvlaovi-flip: I don't think the second icon should be affected, as the one on the left is "primary" and needs the color change in some cases.
I'll check with Liu

matvlaovi-flip: Ok Liu says this is fine, the iconColor will soon be deprecated as it's really only used for the file chips which will get their own component soon

AshleyBekemeier (approved):

PR #1611, opened 2026-04-13
fix: swirl-dialog title and long button labels

fabio-carvalho88:
## Summary

- [Ticket](#)
- [Design](#)
- [Storybook](#)


fabio-carvalho88:
## Changes

If `label` receives a string with `\u00AD` in it, replaces it with an hyphen.
Both buttons will always have the same width, until a maximum of 50% of the parent div.

image

packages/swirl-components/src/components/swirl-dialog/swirl-dialog.css
```diff
@@ -129,23 +133,30 @@
...
+  display: grid;
+  gap: var(--s-space-8);
+  width: fit-content;
+  max-width: 100%;
+  min-width: 0;
+}
+
+.dialog__actions:has(> :nth-child(2)) {
+  grid-template-columns: repeat(2, minmax(0, 1fr));
+}
```
nunofaria11, line 173: **issue:** When **only** a primary or secondary button exists, the padding in the modal seems broken.

Previously vs Now:

Image

Image

packages/swirl-icons/dart/lib/fonts/flip-legacy-icons.ttf
nunofaria11, line 1: **question:** Is there a reason why fonts, icons, and light/dark styles have changed? If they don't need to have changes, we should revert these.

nunofaria11 (changes requested):

packages/swirl-icons/dart/lib/fonts/flip-legacy-icons.ttf
fabio-carvalho88: they were updated when I ran `yarn dev` locally. reverted.

packages/swirl-components/src/components/swirl-dialog/swirl-dialog.css
fabio-carvalho88: fixed

nunofaria11 (approved):

nunofaria11:
LGTM 👍

PR #1614, opened 2026-04-14
Revamp swirl-thumbnail

joseabernardes:
## Summary

- [Ticket](https://linear.app/flip/issue/ENG-3750/improve-thumbnail-component-swirl)
- [Design](https://www.figma.com/design/iRHiAFsHxnzux8gQ9EjlZe/Web-Components?node-id=11456-6052&m=dev)


packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.css
```diff
@@ -105,27 +104,67 @@
...
 
+.thumbnail__compact-action {
+  position: absolute;
+  z-index: 2;
+  top: 0;
+  right: 0;
+}
+
+.thumbnail__compact-button {
+  width: var(--swirl-thumbnail-compact-button-size, 1.25rem);
```
joseabernardes: **note:** I'm especially concerned about the size of this button in touch devices...

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.tsx
```diff
@@ -29,20 +45,34 @@ export class SwirlThumbnail {
...
+    const compactIconSize =
+      this.size === "xl" ? 28 : this.size === "l" ? 24 : 20;
+    const isUploading = this.progress !== undefined;
+
+    const showEditButton = Boolean(this.showEditButton);
+    const showRemoveButton = Boolean(this.showRemoveButton);
+    const hasAnyAction = showEditButton || showRemoveButton;
 
-    const showButtonGroup = showEditButton || showRemoveButton;
+    const showCompactMenu = isCompact && hasAnyAction;
```
Sqrrl: **question:** Should we really show the … menu if only one action is enabled? We could show the action button directly in the corner. That's also how I read the Figma design.

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.tsx
```diff
@@ -1,10 +1,24 @@
...
+} from "@stencil/core";
 import classnames from "classnames";
 
 export type SwirlThumbnailFormat = "portrait" | "landscape" | "square";
 
 export type SwirlThumbnailSize = "s" | "m" | "l" | "xl" | "2xl";
 
+const COMPACT_SIZES: Record<SwirlThumbnailFormat, SwirlThumbnailSize[]> = {
+  landscape: ["s", "m"],
+  square: ["s", "m"],
```
Sqrrl: **issue:** The combination of "square" and size "l" doesn't work. It should also be compact, I guess.

Image

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.css
Sqrrl: Yes. The small three dot menu will not work properly on touch devices.

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.tsx
```diff
@@ -29,20 +45,34 @@ export class SwirlThumbnail {
   @Event() remove: EventEmitter<MouseEvent>;
   @Event() thumbnailClick: EventEmitter<MouseEvent>;
 
-  render() {
-    const showInteractable =
-      (this.size === "xl" || this.size === "2xl") && this.format === "square";
+  // @State so that swirl-popover-trigger re-renders once the ref is set.
+  @State() private popoverEl?: HTMLSwirlPopoverElement;
```
Sqrrl: **issue:** This might be problematic, right? The render is setting the variable, which in turn triggers a re-render, because of `@State`. Can't you use `componentDidLoad`?

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.tsx
joseabernardes: Agreed, I probably forgot to check that combination.

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.css
Sqrrl: If the thumbnail itself is not interactive, we could use the entire thumbnail as the popover trigger, in that case. Or just don't use thumbnail actions on the small sizes.

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.tsx
joseabernardes: Were you worried about an infinite rendering loop, or just double rendering? Still, your idea of setting the `swirlPopover` prop in `componentDidLoad` work :)

Sqrrl: Double render and a Stencil warning during compilation(?). I think Stencil guards against the loop.

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.css
joseabernardes: If we don't show the action on smaller sizes, we'll lose functionality, right?
One alternative would be to add an action item in the popover to "Open" when the thumbnail is interactive.
And then, as you said, the entire thumbnail in the popover trigger.

Sqrrl: Yes. That could work

joseabernardes: Should we implement this for all devices or only for touch devices? While implementing it only for touch devices would at least respect the design of non-touch devices, it would introduce some additional complexity in the component.

We should also consider that the component will now be primarily used in the `m` size, so this will directly affect the app.

Sqrrl: Without the distinction, it would introduce an extra click on non-touch devices if the thumbnail itself is interactive. Not ideal.

joseabernardes: I’ve implemented the distinction. I opted for the approach of always having the entire thumbnail as a popover trigger on touch devices. The compact button would be too small even in the larger thumbnail sizes, and the segmented buttons were only triggered on hover, which would require us to always show them in touch devices, as you did in the resource-list-item.

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.tsx
joseabernardes: Done

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.tsx
joseabernardes: I may have missed the warning, but I didn’t see it. Either way, fixed with the `componentDidLoad` approach.

joseabernardes:
The component ended up being quite complex, but I think it addresses our concerns regarding touch devices.

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.css
Sqrrl: Tested it on my iPhone (Safari), and the bottom sheet is always closing immediately after flashing open briefly.

joseabernardes: Should be fixed https://github.com/getflip/swirl/pull/1614/changes/bab2d1607b01ed60e3eb672b313cbcb03993c56c

The issue was reproducible in all popovers with shadow-dom parents.

joseabernardes: I can still reproduce the issue in other popovers across the app...

joseabernardes: Ended up replacing `body-scroll-lock` with a custom implementation that ensures the body has `width: 100%` on iOS.

Sqrrl: Does that work similarly to the lib? Will it preserve the scroll position for the `overflow: hidden` solution, for example? If it's just about the 100% width, couldn't we do that manually after the bodyScrollLock call?

Sqrrl (approved):

joseabernardes:
@Sqrrl, can you please review the last 5 commits?

Sqrrl (approved):
Lgtm

PR #1616, opened 2026-04-16
File chip revamp

Sqrrl:
## Summary

- [Ticket](https://linear.app/flip/issue/ENG-3706/adjust-file-chip-component)
- [Design](https://www.figma.com/design/JC7QsUEZukSDZytvRjrqIh/UX-Improvements?node-id=2056-36824&m=dev)

### DIFF
--- a/packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.css
+++ b/packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.css
@@ -1,6 +1,11 @@
 :host {
   display: inline-block;
   max-width: 100%;
+  overflow: hidden;
+  background-color: var(--s-background-default);
+  border-radius: var(--s-border-radius-sm);
+  outline: 1px solid var(--s-border-strong);
+  outline-offset: -1px;
 }
 
 .file-chip {
@@ -12,9 +17,6 @@
   font-size: var(--s-font-size-sm);
   line-height: var(--s-line-height-sm);
   background-color: transparent;
-  border-radius: var(--s-border-radius-sm);
-  outline: 1px solid var(--s-border-strong);
-  outline-offset: -1px;
   padding-top: var(--s-space-8);
   padding-right: var(--s-space-16);
   padding-bottom: var(--s-space-8);
@@ -24,18 +26,20 @@
 
 .file-chip:hover:not(.file-chip--no-actions):not(.file-chip--loading),
 .file-chip:focus-within:not(.file-chip--no-actions):not(.file-chip--loading) {
-  &.file-chip--has-preview-action,
-  &.file-chip--has-download-action {
-    background-color: var(--s-state-hovered);
-    cursor: pointer;
-  }
+  @media (--has-hover) {
+    &.file-chip--has-preview-action,
+    &.file-chip--has-download-action {
+      background-color: var(--s-state-hovered);
+      cursor: pointer;
+    }
 
-  & .file-chip__description {
-    visibility: hidden;
-  }
+    & .file-chip__description {
+      visibility: hidden;
+    }
 
-  & .file-chip__actions {
-    visibility: visible;
+    & .file-chip__actions {
+      visibility: visible;
+    }
   }
 }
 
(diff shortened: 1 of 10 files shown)


packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.tsx
```diff
@@ -118,52 +140,90 @@ export class SwirlFileChip {
...
-            )}
+            <span class="file-chip__suffix">
+              {(this.description || this.loading) && (
+                <span class="file-chip__description">
+                  {this.loading ? this.loadingLabel : this.description}
+                </span>
+              )}
+              {hasAction && !this.loading && (
+                <span class="file-chip__actions">
+                  {this.showPreviewButton && (
```
joseabernardes, line 190: **question:** My interpretation of the design spec is that we wouldn't show the preview button anymore. Did you keep it for some reason?

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.tsx
```diff
@@ -118,52 +140,90 @@ export class SwirlFileChip {
...
+                    <swirl-button
+                      disabled={this.loading}
```
joseabernardes: **suggestion:** Disabling the buttons is a bit redundant since we won’t even display them, right? 🙂

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.css
```diff
@@ -1,62 +1,101 @@
 :host {
   display: inline-block;
   max-width: 100%;
+  overflow: hidden;
+  background-color: var(--s-background-default);
```
joseabernardes, line 5: **question:** Should the chip have a background? Or should it be transparent/translucent? I don't see any background in the design spec, that's why I'm asking.

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.css
```diff
@@ -1,62 +1,101 @@
...
+.file-chip__actions {
+  @media (--has-hover) {
+    grid-area: 1 / 1;
+  }
+}
+
 .file-chip__description {
   color: var(--s-text-subdued);
+  font-size: var(--s-font-size-xs);
+  line-height: var(--s-line-height-xs);
```
joseabernardes: **issue:** We see a slight jump when an action is enabled. We could fix it by increasing the `line-height` here to match the height of the actions (24px) 🤔 

I see this as a "issue" because when we’re uploading a file, these actions will be enabled after the initial render, so the jump will be noticeable to the user.

https://github.com/user-attachments/assets/03ef536f-b61e-4664-ab61-65945e1dda03

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.spec.tsx
joseabernardes, line 1: **suggestion:** Can you add a simple test to check that the `delete` event is emitted?

joseabernardes (commented):
This looks much nicer 😍

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.tsx
Sqrrl: Should have mentioned that. Accessibility. I didn't want to make the entire chip a button, because of the nested interactive elements problematic. Was hoping this is okay for Gui. I did go for a compromise by still adding the click handler on the chip.

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.tsx
Sqrrl: Removed.

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.css
Sqrrl: I did it transparent first, but ran into too many issues in Flip Web.

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.css
Sqrrl: Done.

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.spec.tsx
Sqrrl: Of course. Done.

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.tsx
joseabernardes: Yeah, it was challenging enough to fix that in the resource-list-item...
Let's go with it for now.

joseabernardes (approved):

PR #1641, opened 2026-05-05
Fix swirl-image-grid-item reconnect

joseabernardes:
Fix `swirl-image-grid-item` showing the skeleton permanently after a drag-and-drop reorder moves an off-screen item into view. 
The `IntersectionObserver` was only set up in `componentDidLoad` and never recreated on reattach, so `inViewport` stayed `false` and the `` never rendered — now it's also (re)initialized in `connectedCallback`, with a guard to avoid double-setup.

### DIFF
--- a/.changeset/flat-lies-appear.md
+++ b/.changeset/flat-lies-appear.md
@@ -1,0 +1,6 @@
+---
+"@getflip/swirl-components": patch
+---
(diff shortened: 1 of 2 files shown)


Sqrrl (approved):
Nice.

PR #1652, opened 2026-05-07
Add error state to swirl-file-chip

joseabernardes:
## Summary

- [Ticket](https://linear.app/flip/issue/ENG-3933/add-error-state-to-file-chip)

### DIFF
--- a/packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.tsx
+++ b/packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.tsx
@@ -188,7 +188,10 @@ export class SwirlFileChip {
             </span>
             <span class="file-chip__suffix">
               {(this.description || this.loading || this.hasError) && (
-                <span class="file-chip__description">
+                <span
+                  class="file-chip__description"
+                  role={this.hasError ? "status" : undefined}
+                >
                   {this.hasError
                     ? this.errorLabel
                     : this.loading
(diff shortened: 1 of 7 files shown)


packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.tsx
```diff
@@ -37,6 +37,8 @@ export class SwirlFileChip {
   @Prop() deleteButtonLabel?: string = "Delete";
   @Prop() description?: string;
   @Prop() downloadButtonLabel?: string = "Download";
+  @Prop() error?: boolean;
```
Sqrrl: **suggestion:** Just personal preference, and I know we are all over the place with stuff like this in Swirl, but I would prefer `hasError`. It implies the boolean type better.

Sqrrl (approved):

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.tsx
```diff
@@ -180,9 +187,13 @@ export class SwirlFileChip {
               {this.name}
             </span>
             <span class="file-chip__suffix">
-              {(this.description || this.loading) && (
+              {(this.description || this.loading || this.error) && (
                 <span class="file-chip__description">
```
Sqrrl: **note:** I wonder if this should have a status role if it shows an error.

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.tsx
joseabernardes: Makes sense. Let's take the chance that we're introducing the prop now to do it properly :)

packages/swirl-components/src/components/swirl-file-chip/swirl-file-chip.tsx
joseabernardes: Of course!

Sqrrl (approved):

PR #1656, opened 2026-05-08
fix(tokens): increase interactive-primary-disabled opacity in dark mode

marcomFlip:
Raises the dark mode `interactive-primary-disabled` token from `core-translucent-white-6` (6% opacity) to `core-translucent-white-20` (20% opacity) so disabled switches are visible on dark backgrounds.


marcomFlip:
@Sqrrl I was not able to check a preview of my changes, I hope I am not breaking anything. 
Looks that the changes are in the general token, we do not have a dark mode token. 

One test failed, that was the answer from Claude:

>  his is a pre-existing bug in the docs prebuild script — it was broken before our PR and has nothing to do with the token value we changed. You can verify by checking if the same job was failing on main before our branch.

Sqrrl:
Hey @marcomFlip. Changes to the token files cannot be made manually. Everything except the figma-tokens.json file is automatically generated from the Figma tokens. In the past, this was managed directly in Figma via a plugin. Not sure if this is still the case. You would need to ask Liu or Gui about that.

I added your changes to the correct file now.

The token you changed was only for dark mode, correct? It also exists for light mode with a fixed value (#cccdd1). In general, the token appears to be used only by the swirl-switch and swirl-button (with intent "primary", variant "plain", and state "disabled"). Could you verify the change here in light and dark mode, please?

https://icy-water-049ec4003-1656.westeurope.3.azurestaticapps.net/?path=/story/introduction--introduction

marcomFlip:
@Sqrrl yes, only on dark mode. Light mode works perfectly

Sqrrl:
@marcomFlip I forgot about this one. Looks good to you? We can merge this?

Sqrrl (approved):

PR #1668, opened 2026-05-15
Add token tools to Swirl MCP

joseabernardes:


Sqrrl (approved):
I saw that the token list is an array. Agents are supposed to always read the entire list, right? I was wondering if using a map with the token key would have any benefits. It could speed up token lookups and save tokens.

joseabernardes:
> I saw that the token list is an array. Agents are supposed to always read the entire list, right? I was wondering if using a map with the token key would have any benefits. It could speed up token lookups and save tokens.

In theory, the tool's output will be injected into the context, and the agent will not "process" it beyond reading it. Nevertheless, it might make sense to make it a bit slimmer if we see that the context is growing too much after the tool is used.

PR #1675, opened 2026-05-18
Add size and actionsOrientation props to SwirlDialog

marcomFlip:
- Add `size` prop ("default" | "large") — large sets dialog width to 560px on desktop
- Add `actionsOrientation` prop ("horizontal" | "vertical") — uses SwirlButtonGroup in the footer so buttons can be stacked vertically with full width stretch

image


Sqrrl (changes requested):
Some issues/breaking changes here:

**1. The buttons in the default variant no longer span the entire width on small viewports.**

Before:
Image

After:
Image

**2. The buttons no longer truncate correctly in default mode.**

Before:
Image

After:
Image

marcomFlip:
> 1. The buttons in the default variant no longer span the entire width on small viewports.
Also fixed

image

> 2. The buttons no longer truncate correctly in default mode.
Also fixed


image

thanks for checking @Sqrrl

Sqrrl (approved):

PR #1678, opened 2026-05-20
Fix SwirlSelect indicator pointer events

Sqrrl:


joseabernardes:
**question:** How can I reproduce the issue?

Sqrrl:
> **question:** How can I reproduce the issue?

You can't, really. 😬 It only shows when you have the select connected to a Quill editor, and that editor has a text range selected. In other cases, the click on the icon works by accident because of how it works inside the swirl-form-control.

joseabernardes:
> > **question:** How can I reproduce the issue?
> 
> You can't, really. 😬 It only shows when you have the select connected to a Quill editor, and that editor has a text range selected. In other cases, the click on the icon works by accident because of how it works inside the swirl-form-control.

Great instructions - that helped me reproduce it in your MR 😄

joseabernardes (approved):

PR #1680, opened 2026-05-22
Added fixed aspect ratio prop

AshleyBekemeier:
## Summary

Added a prop to allow the Date picker to be stretched wider without taking up insane height.


matvlaovi-flip (approved):
I think it's a nice change.
Have you verified our current usages in Web/Console if it has some unintended consequences?

PR #1697, opened 2026-06-08
fix(components): fix pointer events issue with SwirlPopover

Sqrrl:
## Summary

The issue is caused by popovers being nested inside swirl-resource-list-item controls, which have disabled pointer events when not hovered/focused.

This is more of a workaround. Technically, swirl-popover elements shouldn't be nested inside a swirl-resource-list-item "control" slot, but sometimes our Angular component structure makes it hard to avoid.

- [Ticket](https://linear.app/flip/issue/ENG-4158/chat-popover-actions-dont-work)

### DIFF
--- a/.changeset/good-kings-pump.md
+++ b/.changeset/good-kings-pump.md
@@ -1,0 +1,6 @@
+---
+"@getflip/swirl-components": patch
+---
+
+Fix swirl-popover pointer events when nested inside swirl-resource-list-item's
+control slot
--- a/packages/swirl-components/src/styles/global.css
+++ b/packages/swirl-components/src/styles/global.css
@@ -126,6 +126,17 @@ dt {
   padding: 0;
 }
 
+/*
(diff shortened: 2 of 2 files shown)


nunofaria11 (approved):

joseabernardes:
**question:** Why was this not an issue before? 🫣

Sqrrl:
> **question:** Why was this not an issue before? 🫣

Maybe something changed about how Chromium evaluates the hover/focus. Firefox still doesn't have the issue. Maybe the focus/hover inside the top-layer is no longer treated as focus/hover of the resource list item.

Sqrrl:
6 days ago: [https://developer.chrome.com/release-notes/149?hl=en#user_action_pseudo-class_top_layer_boundary](<https://developer.chrome.com/release-notes/149?hl=en#user_action_pseudo-class_top_layer_boundary>)

PR #1709, opened 2026-06-12
Fix swirl thumbnail remove prop collision

danizep:
## Summary

Rename `remove` event to `removeThumbnail` to avoid conflicts with native `HTMLElement.remove` function

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### DIFF
--- a/packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.tsx
+++ b/packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.tsx
@@ -44,7 +44,7 @@ export class SwirlThumbnail {
   @Prop() timestamp?: string;
 
   @Event() edit: EventEmitter<MouseEvent>;
-  @Event() remove: EventEmitter<MouseEvent>;
+  @Event({ eventName: "remove" }) removeThumbnail?: EventEmitter<MouseEvent>;
   @Event() thumbnailClick: EventEmitter<MouseEvent>;
 
   @State() hasHover: boolean = true;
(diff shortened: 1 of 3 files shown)


packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.tsx
```diff
@@ -44,7 +44,7 @@ export class SwirlThumbnail {
   @Prop() timestamp?: string;
 
   @Event() edit: EventEmitter<MouseEvent>;
-  @Event() remove: EventEmitter<MouseEvent>;
```
Sqrrl, line 47: **question:** This way it could still be confused with e.g., a bubbling native remove event, right? I wonder if we should make this a breaking change and rename the event entirely. What do you think?

Sqrrl (approved):

packages/swirl-components/src/components/swirl-thumbnail/swirl-thumbnail.tsx
Sqrrl: Nvm. Just realized that there is no native "remove" event. 😅

danizep: Yeah, I would be okay with it, but i'm currently a bit busy.
The same pattern applies to Chips but is working properly because implements these alias.
It might also apply to other components. 

If you are ok, I would treat these as technical debt and take care of it later. 
On the client the problem was mitigated by wrapping the component with the HTML element, this change is more of a sanity check to make sure we have no regressions if somebody removes a wrapping element.

Sqrrl: Sorry, I got confused. See my second comment. I don't think we have to change the event name.

> Nvm. Just realized that there is no native "remove" event. 😅

PR #1720, opened 2026-06-18
fix: data cell intent background colors

fabio-carvalho88:
## Summary

- [Ticket](https://linear.app/flip/issue/ACT-3579/swirl-fix-data-cell-intent-background-colors)
- [Design](#)
- [Storybook](#)

If necessary you can **[check here](https://www.figma.com/design/zI4W3OiXq6Lnr7QwYcNaFB/Paswordless-Login-via-Passkeys?node-id=868-42368&m=dev)** the cells background colors.
Take SSO as an example, the parent cell uses a lighter grey `var(--s-surface-raised-50-default)` and the children use a darker one with `var(--s-surface-raised-default)`

https://github.com/user-attachments/assets/4cf3a35b-705a-4e6d-b398-82cbd6688cc9

### DIFF
--- a/.changeset/five-steaks-protect.md
+++ b/.changeset/five-steaks-protect.md
@@ -1,0 +1,5 @@
+---
+"@getflip/swirl-components": minor
+---
+
+Fix data cell intent background colors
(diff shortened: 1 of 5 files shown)


flipandi (approved):
Looks good

PR #1732, opened 2026-06-23
Upgrade deps

Sqrrl:
Upgraded some vulnerable dependencies, including major upgrades for turbo and uuid.

### DIFF
--- a/packages/bridge/package.json
+++ b/packages/bridge/package.json
@@ -30,7 +30,6 @@
     "@babel/preset-typescript": "7.18.6",
     "@rollup/plugin-typescript": "10.0.1",
     "@types/jest": "27.5.2",
-    "@types/uuid": "8.3.4",
     "babel-jest": "29.7.0",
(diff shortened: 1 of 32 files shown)


packages/bridge/package.json
```diff
@@ -23,7 +23,7 @@
   },
   "dependencies": {
     "@getflip/swirl-components": "0.507.0",
-    "uuid": "9.0.1"
+    "uuid": "14.0.1"
```
joseabernardes: **suggestion:** What if we ditch `uuid` in favor of `crypto.randomUUID`?

joseabernardes (approved):

danizep (approved):
I was just waiting to seen if getflip.dev build would pass on Vercel. Looks good.

packages/bridge/package.json
Sqrrl: Good idea. Done.

PR #1751, opened 2026-07-21
style: add min-width to swirl-button for improved layout consistency

fabio-carvalho88:
## Summary

- [Ticket](https://linear.app/flip/issue/ACT-1112/swirl-set-min-width-of-swirl-button-to-line-height)

### DIFF
--- a/packages/swirl-components/src/components/swirl-button/swirl-button.css
+++ b/packages/swirl-components/src/components/swirl-button/swirl-button.css
@@ -26,6 +26,7 @@
   font: inherit;
   font-weight: var(--s-font-weight-medium);
   line-height: var(--s-line-height-lg);
+  min-width: var(--s-line-height-lg);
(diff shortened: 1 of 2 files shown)


packages/swirl-components/src/components/swirl-button/swirl-button.css
flipandi, line 481: **question (non-blocking)** should we also do something here that ensures we have at least a circle in width?

fabio-carvalho88: i don’t think we need anything specific here. the radius already produce the capsule shape, and min-width is set from line-height on the base button, which pills inherit

flipandi:
LGTM

flipandi (approved):

flipandi (approved):

PR #1753, opened 2026-07-21
fix(swirl-button): use strong text/icon color tokens for intent=strong + variant=outline

marcomFlip:
When a SwirlButton has intent=strong and variant=outline, the icon and label render with the subdued/default gray tokens instead of the intended strong/default dark tokens. The button looks muted rather than strong.
This pr will fix that.
Affected condition (only this combination): intent="strong" AND variant="outline"

### DIFF
--- a/.changeset/swirlbutton-outline-strong-colors.md
+++ b/.changeset/swirlbutton-outline-strong-colors.md
@@ -1,0 +1,5 @@
+---
+"@getflip/swirl-components": patch
+---
+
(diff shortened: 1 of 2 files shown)


matvlaovi-flip:
PR is missing the changeset, do you know how to do that?

marcomFlip:
> PR is missing the changeset, do you know how to do that?

@matvlaovi-flip - should be there now, I asked claude 🤣

matvlaovi-flip (approved):

PR #1756, opened 2026-07-28
Fix SwirlProgressIndicator (variant=bar) size heights to and add a new l size for both variants 

marcomFlip:
## Summary

`swirl-progress-indicator`'s `bar` variant rendered `xs` and `s` at the same height (2px) and had no `l` size. This updates the CSS to match the desired size mapping and adds the missing `l` size.

**Before → after (bar height)**

| size | before | after |
|------|--------|-------|
| xs   | 2px    | 1px   |
| s    | 2px    | 2px (unchanged) |
| m    | 4px    | 4px (unchanged, still default) |
| l    | n/a    | 8px (new) |

Since `size` is shared with the `circle` variant, and `circleSizeConfig` is a `Record<SwirlProgressIndicatorSize, ...>` requiring every key, I added a proportional `l` entry for `circle` too (`radius: 40, strokeWidth: 8` — double the `m` config, matching the existing `s`→`m` stroke/diameter ratio). No spec called out circle geometry for `l`, so flagging this choice for review.

## What a reviewer should know

- `custom-elements.manifest.json` and `vscode-data.json` are regenerated build artifacts (via `stencil build`) reflecting the new `l` size in the type union — not hand-edited.
- Added spec coverage for the new `l` size on both `bar` and `circle` variants.
- Added a changeset (`patch`).

## Test plan

- [x] `stencil test --spec` — all 508 tests pass (including new `l`-size cases)
- [x] `eslint` on changed files — clean
- [x] Verified in a local Storybook build: `xs`/`s`/`m`/`l` bar heights measured via computed style match 1px/2px/4px/8px; unset `size` still resolves to the `m` default (4px)

### DIFF
--- a/.changeset/fix-progress-bar-sizes.md
+++ b/.changeset/fix-progress-bar-sizes.md
@@ -1,0 +1,5 @@
+---
+"@getflip/swirl-components": patch
+---
+
+Fix SwirlProgressIndicator (variant=bar) size heights to match the design spec (xs: 1px, s: 2px, m: 4px) and add the missing l size (8px)
--- a/.changeset/fix-progress-bar-sizes.md
+++ b/.changeset/fix-progress-bar-sizes.md
@@ -1,5 +1,5 @@
 ---
-"@getflip/swirl-components": patch
+"@getflip/swirl-components": minor
 ---
 
 Fix SwirlProgressIndicator (variant=bar) size heights to match the design spec (xs: 1px, s: 2px, m: 4px) and add the missing l size (8px)
--- a/packages/swirl-components/custom-elements.manifest.json
+++ b/packages/swirl-components/custom-elements.manifest.json
@@ -46941,7 +46941,7 @@
             {
               "name": "size",
               "type": {
-                "text": "\"m\" | \"s\" | \"xs\"",
+                "text": "\"l\" | \"m\" | \"s\" | \"xs\"",
                 "references": [
                   {
                     "name": "SwirlProgressIndicatorSize"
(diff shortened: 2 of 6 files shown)


.changeset/fix-progress-bar-sizes.md
fabio-carvalho88, line 1: this change adds a new public value (l) to SwirlProgressIndicatorSize, our convention reserves `patch` for things like bugfixes, additive-API changes should ship as `minor`.

on top of that claude claims the following:

> .changeset/fix-progress-bar-sizes.md:5 — description doesn't match the diff
> 
>   The entry scopes itself to variant=bar, but the commit also adds a circle l: l: { radius: 40, strokeWidth: 8 } (swirl-progress-indicator.tsx:15) and a 96×96 box (swirl-progress-indicator.css:113-116). The
>   circle addition is necessary — circleSizeConfig is a Record<SwirlProgressIndicatorSize, …> and would fail to typecheck otherwise — so the code is right; the changelog just hides it. Anyone reading the
>   release notes won't learn a 6rem circle now exists. **Suggested wording:**
> 
>   **Fix SwirlProgressIndicator (variant=bar) size heights to match the design spec (xs: 1px, s: 2px, m: 4px) and add a new l size for both variants (bar: 8px, circle: 96px)**

fabio-carvalho88 (changes requested):

.changeset/fix-progress-bar-sizes.md
marcomFlip: @fabio-carvalho88 thanks, changed the title only, as the description and summary seems good to me

fabio-carvalho88 (approved):

PR #1757, opened 2026-07-30
Tree view custom icons

danizep:
## Summary

SwirlTreeView

- Remove drag handle
- Support custom expand and collapse icons
- Add grouping line on child items

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### DIFF
--- a/packages/swirl-components/src/components/swirl-tree-view-item/swirl-tree-view-item.tsx
+++ b/packages/swirl-components/src/components/swirl-tree-view-item/swirl-tree-view-item.tsx
@@ -545,11 +545,6 @@ export class SwirlTreeViewItem {
             role={semantics === "tree" ? "treeitem" : undefined}
             tabIndex={tabIndex}
           >
-            {!this.disableDrag && this.enableDragDrop && (
-              <span class="tree-view-item__drag-handle">
-                <swirl-icon-drag-handle size={20}></swirl-icon-drag-handle>
-              </span>
-            )}
             {this.expandable && semantics === "tree" && (
               <span class="tree-view-item__toggle-icon">
                 {hasChildren && (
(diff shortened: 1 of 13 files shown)


packages/swirl-components/src/components/swirl-tree-view-item/swirl-tree-view-item.tsx
```diff
@@ -528,25 +545,22 @@ export class SwirlTreeViewItem {
             role={semantics === "tree" ? "treeitem" : undefined}
             tabIndex={tabIndex}
           >
-            {!this.disableDrag && this.enableDragDrop && (
-              <span class="tree-view-item__drag-handle">
-                <swirl-icon-drag-handle size={20}></swirl-icon-drag-handle>
-              </span>
-            )}
```
Sqrrl, line 535: **issue:** I don't love removing the drag handle. It serves as an important visual indicator for the drag-and-drop functionality.

packages/swirl-components/src/components/swirl-tree-view/swirl-tree-view.css
Sqrrl, line 1: **suggestion:** I realized that the preview in Storybook doesn't look correct (https://icy-water-049ec4003-1757.westeurope.3.azurestaticapps.net/). There are some global Storybook styles leaking into the component styles. Maybe you could fix that, so the preview actually looks like it will in our apps.

Image

Same for the underline on hover:

Image

Maybe a `:not(.docs-story)` in the selectors of the global styles would help.

danizep: Yeah, I noticed that I ignore it because it looks good on the playground, but yes I will do it.

packages/swirl-components/src/components/swirl-tree-view-item/swirl-tree-view-item.tsx
danizep: I know, but this is a design decision in favor of introducing the guiding lines.

We can see platforms like Notion and Linear where the exact same functionality is supported without affordances that it can be dragged. The assumption is that these patterns are well known and the user will just try it and do it.

Furthermore, after discussing with design we will have a follow-up ticket to replace the illustrative icon with an arrow for expand/collapse on `:hover` so alignment is better and we save even more space horizontally. 

Screenshot 2026-07-31 at 11 46 28

Sqrrl (approved):

PR #1762, opened 2026-08-05
Add swirl-illustration-* component generation

Sqrrl:
## Summary

`yarn generate` to generate the components from the illustration SVG assets. The new components automatically work with light and dark mode. After merging, we can start replacing the inline SVGs in our apps with the new component(s).

- [Ticket](https://linear.app/flip/issue/COR-1191/add-empty-illustrations-to-swirlemptystate-in-storybook)
- [Storybook](https://icy-water-049ec4003-1762.westeurope.3.azurestaticapps.net/?path=/docs/components-swirlillustration--docs)


joseabernardes:
**thought:** I know we already do this for icons and symbols, but creating a brand new component for every single illustration feels like too much. IMO, it inflates our component catalog without really reflecting the number of components we actually have.
Would a single generic `swirl-illustration` component (taking the asset name as a prop) work? Or is it done this way because we need to embed the SVG directly in the component?
Again, food for thought, and because I never shared my opinion on this pattern before, and I also don't know yours :)

Sqrrl:
> **thought:** I know we already do this for icons and symbols, but creating a brand new component for every single illustration feels like too much. IMO, it inflates our component catalog without really reflecting the number of components we actually have. Would a single generic `swirl-illustration` component (taking the asset name as a prop) work? Or is it done this way because we need to embed the SVG directly in the component? Again, food for thought, and because I never shared my opinion on this pattern before, and I also don't know yours :)

I wanted to avoid having a really big component with all the SVGs inlined in it. We can't go for using the SVG assets directly because of light/dark mode. (Technically we could, but only by detecting theme changes in JS, which isn't great.)

What are your concerns about the component count? As you said, we already do it for symbols and icons. In Storybook we show those components as just one.

joseabernardes:
> > **thought:** I know we already do this for icons and symbols, but creating a brand new component for every single illustration feels like too much. IMO, it inflates our component catalog without really reflecting the number of components we actually have. Would a single generic `swirl-illustration` component (taking the asset name as a prop) work? Or is it done this way because we need to embed the SVG directly in the component? Again, food for thought, and because I never shared my opinion on this pattern before, and I also don't know yours :)
> 
> I wanted to avoid having a really big component with all the SVGs inlined in it. We can't go for using the SVG assets directly because of light/dark mode. (Technically we could, but only by detecting theme changes in JS, which isn't great.)
> 
> What are your concerns about the component count? As you said, we already do it for symbols and icons. In Storybook we show those components as just one.

My thought came mainly from when I was building the MCP, because I was forced to create separate tools for these components to avoid cluttering the component list with an endless list of `swirl-icon-X`s. But that issue is solved, I know.  I guess it's just my opinion around having 1 vs N components, as for me this is just 1 component, with N "variants".

Let's go as is.

joseabernardes (approved):

Sqrrl:
> > > **thought:** I know we already do this for icons and symbols, but creating a brand new component for every single illustration feels like too much. IMO, it inflates our component catalog without really reflecting the number of components we actually have. Would a single generic `swirl-illustration` component (taking the asset name as a prop) work? Or is it done this way because we need to embed the SVG directly in the component? Again, food for thought, and because I never shared my opinion on this pattern before, and I also don't know yours :)
> > 
> > 
> > I wanted to avoid having a really big component with all the SVGs inlined in it. We can't go for using the SVG assets directly because of light/dark mode. (Technically we could, but only by detecting theme changes in JS, which isn't great.)
> > What are your concerns about the component count? As you said, we already do it for symbols and icons. In Storybook we show those components as just one.
> 
> My thought came mainly from when I was building the MCP, because I was forced to create separate tools for these components to avoid cluttering the component list with an endless list of `swirl-icon-X`s. But that issue is solved, I know. I guess it's just my opinion around having 1 vs N components, as for me this is just 1 component, with N "variants".
> 
> Let's go as is.

I see. Didn't consider the MCP. I think it's not uncommon, though. I think most big UI libraries do the same. You end up with huge components otherwise. And they are not really tree-shakeable.

PR #1771, opened 2026-08-19
Add progressive selection mode to resource list item

joseabernardes:
## Summary

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### DIFF
--- a/packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
+++ b/packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
@@ -7,6 +7,7 @@ import {
   EventEmitter,
   h,
   Host,
+  Listen,
   Prop,
   State,
 } from "@stencil/core";
@@ -17,6 +18,7 @@ import {
   SwirlResourceListItemAriaCurrent,
   SwirlResourceListItemLabelWeight,
   SwirlResourceListItemRel,
+  SwirlResourceListItemSelectionMode,
   SwirlResourceListItemTarget,
 } from "./swirl-resource-list-item.types";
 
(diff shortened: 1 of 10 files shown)


packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
```diff
@@ -321,15 +387,33 @@ export class SwirlResourceListItem {
...
           </Tag>
-          {this.selectable && (
-            <span aria-hidden="true" class="resource-list-item__checkbox">
-              <span class="resource-list-item__checkbox-icon">
-                {this.checked && (
-                  <swirl-icon-check-strong></swirl-icon-check-strong>
-                )}
+          {this.selectable &&
+            (hasInteractiveCheckbox ? (
+              <button
```
Sqrrl, line 372: **question:** Is this a button for styling reasons?

Sqrrl (commented):
https://github.com/user-attachments/assets/c892ad1b-2ff1-48a0-a1b7-442648f54e31

There is an issue with tabbing. After a checkbox is selected and I use the arrow keys, the selected checkbox keeps being reachable via Tab even though the parent item is no longer selected.

packages/swirl-components/src/components/swirl-resource-list-item/swirl-resource-list-item.tsx
joseabernardes: Styling and consistency with the current approach, as the current selectable resource-list-item is also a button with the `checkbox` role.

joseabernardes:
> Screen.Recording.2026-08-20.at.09.50.52.mov 
> There is an issue with tabbing. After a checkbox is selected and I use the arrow keys, the selected checkbox keeps being reachable via Tab even though the parent item is no longer selected.

Good catch! I thought I'd covered that, but I made the rookie mistake of forgetting that focus/blur events don't bubble :/
Should be fixed now.

Sqrrl (approved):

PR #1777, opened 2026-08-25
feat: support a status badge on swirl-tab-bar tabs

fabio-carvalho88:
## Summary

- [Ticket](https://linear.app/flip/issue/ACT-4193/swirl-support-a-status-badge-on-swirl-tab-bar-tabs)
- [Design](https://www.figma.com/design/LWIJbG2TxWy86KMsupxTl2/Bulk-Invite-Code-Generation?node-id=2133-88125&m=dev)

### DIFF
--- a/packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.css
+++ b/packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.css
@@ -186,5 +186,5 @@
 }
 
 .tab-bar__tab-badge {
-  padding-left: var(--s-space-4);
+  padding-left: var(--s-space-6);
 }
--- a/packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.css
+++ b/packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.css
@@ -150,6 +150,11 @@
     & .tab-bar__tab-suffix {
       color: var(--s-text-subdued);
     }
+
+    /* the badge sits inline on a pill, so it needs no border to separate it from its surroundings */
+    & .tab-bar__tab-badge {
+      --swirl-badge-border-color: transparent;
+    }
   }
 
   @media (--from-desktop-without-touch) {
@@ -178,4 +183,8 @@
 .tab-bar__tab-suffix {
   color: var(--s-interactive-neutral-default);
   font-weight: var(--s-font-weight-normal);
+}
+
+.tab-bar__tab-badge {
+  padding-left: var(--s-space-4);
 }
--- a/packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.css
+++ b/packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.css
@@ -186,5 +186,5 @@
 }
 
 .tab-bar__tab-badge {
-  padding-left: var(--s-space-6);
+  padding-left: calc(var(--s-space-4) + var(--s-space-2));
 }
--- a/packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.stories.ts
+++ b/packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.stories.ts
@@ -1,4 +1,5 @@
 import { generateStoryElement } from "../../utils";
+import { SwirlTabBarTab } from "./swirl-tab-bar";
 import Docs from "./swirl-tab-bar.mdx";
 
 export default {
@@ -18,29 +19,44 @@ const Template = (args) => {
   return element;
 };
 
+// the badge sits on the active tab so the pill variant story shows the
+// combination where the intent color has no contrast and has to be adjusted
+const tabs: SwirlTabBarTab[] = [
+  {
+    active: false,
+    id: "tab1",
+    label: "A Tab",
+    suffix: "2",
+    tooltip: "With a tooltip",
+  },
+  {
+    active: true,
+    badge: {
+      intent: "info",
+      label: "Batches running",
+    },
+    icon: '<swirl-icon glyph="emoji-mood" size="20"></swirl-icon>',
+    id: "tab2",
+    label: "Another Tab",
+    suffix: "12",
+  },
+  {
+    active: false,
+    icon: '<swirl-icon glyph="emoji-satisfied" size="20"></swirl-icon>',
+    id: "tab3",
+    label: "Yet Another Tab",
+  },
+];
+
 export const SwirlTabBar = Template.bind({});
 
 SwirlTabBar.args = {
-  tabs: [
-    {
-      active: false,
-      id: "tab1",
-      label: "A Tab",
-      suffix: "2",
-      tooltip: "With a tooltip",
-    },
-    {
-      active: true,
-      icon: '<swirl-icon glyph="emoji-mood" size="20"></swirl-icon>',
-      id: "tab2",
-      label: "Another Tab",
-      suffix: "12",
-    },
-    {
-      active: false,
-      icon: '<swirl-icon glyph="emoji-satisfied" size="20"></swirl-icon>',
-      id: "tab3",
-      label: "Yet Another Tab",
-    },
-  ],
+  tabs,
+};
+
+export const WithPillVariant = Template.bind({});
+
+WithPillVariant.args = {
+  tabs,
+  variant: "pill",
 };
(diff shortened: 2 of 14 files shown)


fabio-carvalho88:
### Why changing a tab prop should update the strip

`swirl-tabs` builds its strip by reading `icon/label/tooltip` off its child swirl-tab elements. On main, updateTabBarTabs() does that in exactly two places —`componentWillLoad` and the end of `activateTab()` — but it re-reads them live each time. So a changed prop wasn't lost forever; it was picked up on the next tab switch.

The old behavior was therefore: change label, nothing happens; the user clicks a different tab, and now the label updates. Stale, then silently correct. (activateTab() early-returns when the id is unchanged, so re-clicking the same tab didn't do it.)

This PR emits a tabStripPropChange event from a `@Watch` on badge, icon, label and tooltip, which swirl-tabs listens for — so the strip updates when the prop actually changes.

Why it's worth doing for all four rather than just badge:

- label, icon and tooltip are usually set once from i18n and rarely change, so nobody hit this. The one realistic case is a **language switch**, where the strip keeps the old language until the user happens to click another tab.

- badge (new here) is the opposite: it's driven by background state, not navigation. Nobody clicks a tab because a batch started, so a badge could sit invisible indefinitely. That's what forced the event — broadening it to the other three closes the same hole for them.

packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.css
```diff
@@ -179,3 +184,7 @@
   color: var(--s-interactive-neutral-default);
   font-weight: var(--s-font-weight-normal);
 }
+
+.tab-bar__tab-badge {
+  padding-left: var(--s-space-4);
```
joseabernardes: **question:** The badge position doesn't seem to be aligned with Figma. Is that intended?

Image

packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.tsx
```diff
@@ -139,6 +151,15 @@ export class SwirlTabBar {
               "tab-bar__tab-label--variant-pill": this.variant === "pill",
             });
 
+            // a dot has no text, so its intent color is the only thing carrying
+            // its meaning — and no intent reads against the active pill's filled
+            // primary background, `info` resolving to the pill color itself and
+            // leaving nothing visible. `neutral` is legible there instead.
+            const badgeIntent: SwirlBadgeIntent =
```
joseabernardes: **thought:** I understand that having something other than "neutral" would not look good with the pill variant, but I would rather leave that responsibility to the "user" of the component and not have it as a strict rule here. Just a general thought.

packages/swirl-components/src/components/swirl-tab/swirl-tab.tsx
```diff
@@ -10,6 +11,9 @@ export type SwirlTabPadding = "0" | "2" | "4" | "8" | "12" | "16" | "20" | "24";
 })
 export class SwirlTab {
   @Prop() active?: boolean;
+  // `badge`, `icon`, `label` and `tooltip` style the tabs
```
joseabernardes: **nitpick:** I would remove the comment tbh.

packages/swirl-components/src/components/swirl-tabs/swirl-tabs.tsx
```diff
@@ -117,9 +117,13 @@ export class SwirlTabs {
     );
   }
 
+  // snapshots the tab item props of every swirl-tab into the state the tab bar
```
joseabernardes: **nitpick:** Same as I wrote in other places. I would skip comments when they don't convey information the reader can get from the code itself.

packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.stories.ts
```diff
@@ -18,29 +19,44 @@ const Template = (args) => {
...
-      active: false,
-      icon: '<swirl-icon glyph="emoji-satisfied" size="20"></swirl-icon>',
-      id: "tab3",
-      label: "Yet Another Tab",
-    },
-  ],
+  tabs,
+};
+
+export const WithPillVariant = Template.bind({});
```
joseabernardes: **thought:** I would skip these additional stories, because stories add value when we can't preview something just by ticking the props. In this case, we can just change the variant in the main stories and achieve the same result.

packages/swirl-components/src/components/swirl-tab/swirl-tab.mdx
```diff
@@ -25,6 +25,27 @@ The SwirlTab component is used to display a tab panel inside the
...
 />
 
+## Tab item props
+
+`swirl-tab` renders only the tab panel. The `badge`, `icon`, `label` and
+`tooltip` props describe the corresponding tab bar item, and are forwarded by
+[SwirlTabs](?path=/docs/components-swirltabs--docs) to
+[SwirlTabBar](?path=/docs/components-swirltabbar--docs).
+
+`badge` renders a status badge — a dot — after the tab label. Because it is an
```
joseabernardes: **praise:** I like the additional documentation, it gives a bit more context on how to build a tab layout (using swirl-tab + swirl-tab-bar). 

**suggestion:** That said, I would remove this paragraph on `badge`, as it's already well documented in the `swirl-tab-bar` docs.

packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.css
fabio-carvalho88: figma isn't consistent with the badge position, some examples:

image

image

fabio-carvalho88: with that said, i'll increase the padding to 6

packages/swirl-components/src/components/swirl-tabs/swirl-tabs.tsx
fabio-carvalho88: what I actually wanted to flag is a current limitation that isn’t obvious from this function. We only snapshot on load and on tab click, so a prop change at runtime (e.g. badge with a counter) won’t show up until the next click. That’s fine for now because we won't be supporting badge with counters, but that was something that was discuss as a possibility during design and implementation, hence why I wanted it visible if we later need live updates.

i'm more than ok with removing the comment, but wdyt about leaving a small note about that?

packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.tsx
fabio-carvalho88: Yeah, I went back and forth on this while implementing it. Forcing the neutral intent on the active pill hides the contrast problem the user (maybe) should actually see and decide on... variable removed ✅

packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.stories.ts
fabio-carvalho88: done ✅

packages/swirl-components/src/components/swirl-tab/swirl-tab.mdx
fabio-carvalho88: done ✅

packages/swirl-components/src/components/swirl-tab/swirl-tab.tsx
fabio-carvalho88: done ✅

packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.css
joseabernardes: You think you did, but you didn't (`--s-space-6` is not a valid var) 😆

packages/swirl-components/src/components/swirl-tabs/swirl-tabs.tsx
joseabernardes: I see your point. Then it makes sense to leave a small note, I agree.

packages/swirl-components/src/components/swirl-tab-bar/swirl-tab-bar.css
fabio-carvalho88: lol nice one...

joseabernardes (approved):

PR #1781, opened 2026-08-27
Stop modelClose and modelOpen event bubbling of SwirlPdfReader

Sqrrl:


joseabernardes (approved):
Let's make sure we don't rely on the bubbling in any place in our apps :)

Sqrrl:
> Let's make sure we don't rely on the bubbling in any place in our apps :)

Already did. We don't rely on it.

PR #1783, opened 2026-08-28
Add interactive prop to swirl-tag

matvlaovi-flip:
## Summary

- [Ticket](https://linear.app/flip/issue/FUS-885/update-swirl-tag-to-have-an-interactive-property)

### DIFF
--- a/packages/swirl-components/src/components/swirl-tag/swirl-tag.css
+++ b/packages/swirl-components/src/components/swirl-tag/swirl-tag.css
@@ -202,6 +202,39 @@
   }
 }
 
+.tag--interactive {
+  position: relative;
+  margin: 0;
+  font-family: inherit;
+  text-align: start;
+  cursor: pointer;
+  appearance: none;
+
+  &::after {
+    position: absolute;
+    border-radius: inherit;
+    content: "";
+    inset: calc(-1 * var(--s-border-width-default));
+    pointer-events: none;
+  }
+
+  &:hover::after {
+    background-color: var(--s-state-hovered);
+  }
+
+  &:active::after {
+    background-color: var(--s-state-pressed);
+  }
+
+  &:focus:not(:focus-visible) {
+    outline: none;
+  }
+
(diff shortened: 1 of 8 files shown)


packages/swirl-components/src/components/swirl-tag/swirl-tag.css
```diff
@@ -202,6 +202,39 @@
   }
 }
 
+.tag--interactive {
+  position: relative;
+  margin: 0;
+  font-family: inherit;
+  text-align: start;
```
fabio-carvalho88: nitpick: this seems redundant as the tag class is already `inline-flex`

packages/swirl-components/src/components/swirl-tag/swirl-tag.css
```diff
@@ -202,6 +202,39 @@
   }
 }
 
+.tag--interactive {
+  position: relative;
+  margin: 0;
```
fabio-carvalho88: question: i've removed this and didn't saw any diff. however i wonder if we need it 🤔

packages/swirl-components/src/components/swirl-tag/swirl-tag.css
```diff
@@ -202,6 +202,39 @@
...
+  font-family: inherit;
+  text-align: start;
+  cursor: pointer;
+  appearance: none;
```
fabio-carvalho88: nitpick: this can be removed as we already control the appearance is css rules above (e.g. padding and background)

fabio-carvalho88 (changes requested):

packages/swirl-components/src/components/swirl-tag/swirl-tag.css
matvlaovi-flip: Thanks, good catch! I removed all 3 as they were unnecessary

fabio-carvalho88 (approved):

PR #1785, opened 2026-09-01
Add mobileBottomSheet prop to swirl-popover

joseabernardes:
## Summary

The popover always opened as a bottom sheet below 768px, which is awkward for popovers opened on hover.

This PR added a `mobileBottomSheet` property that defaults to true. Pass false to keep the popover anchored to its trigger on all viewports.

### DIFF
--- a/packages/swirl-components/src/components/swirl-popover/swirl-popover.css
+++ b/packages/swirl-components/src/components/swirl-popover/swirl-popover.css
@@ -30,11 +30,7 @@
   }
 
   & .popover__content {
-    animation: 0.15s popover-slide-in;
-
-    @media (--from-tablet) {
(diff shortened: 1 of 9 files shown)


packages/swirl-components/src/components/swirl-popover/swirl-popover.css
joseabernardes, line 1: **note:** The diff seems big, but it's mainly moving the specific bottom-sheet styles to an isolated block at the end, instead of being spread across the entire file with all the `@media` queries...

danizep: Cool

danizep (approved):

PR #1787, opened 2026-09-02
SWR-122 Fix swirl-form-control's inline prop rendering an overlapping label

marcomFlip:
## Summary

- [Storybook](https://swirl-storybook.flip-app.dev/?path=/docs/components-swirltimeinput--docs)

`swirl-form-control`'s `inline` prop had its own broken label CSS: it faded the floating label to `opacity: 0`, but only when the child input has no `placeholder`. `swirl-time-input` always sets a default placeholder (`hh:mm`), so the label never faded and rendered on top of the input text.

`labelPosition="outside"` already implements the correct visual (label above, outside the border) — the same layout the ticket asks for `inline` to produce. This PR normalizes `inline` to render with that same effective label position instead of maintaining a second, divergent (and broken) set of label styles, and removes the now-dead CSS block.

Since the fix lives in the shared `swirl-form-control`, it also resolves the same latent bug on any other input with a default placeholder (e.g. `swirl-date-input`), with no changes needed there.

## Test plan

- [x] Added a regression test asserting `inline` renders `form-control--label-position-outside`
- [x] Full component test suite passes (537/537)
- [x] Package type-check passes
- [x] Verified visually in Storybook: `swirl-time-input` with `inline=true` now renders the label cleanly above the border, no overlap
- [x] Verified no regression on `swirl-text-input` (no default placeholder) in inline mode

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/tame-donkeys-hunt.md
+++ b/.changeset/tame-donkeys-hunt.md
@@ -1,0 +1,5 @@
+---
+"@getflip/swirl-components": patch
+---
+
+Fix `swirl-form-control`'s `inline` prop rendering an overlapping floating label instead of positioning the label outside the border. `inline` now renders the same layout as `labelPosition="outside"`, which fixes the broken text alignment on `swirl-time-input` (and any other input with a default placeholder, e.g. `swirl-date-input`) when used inline.
--- a/.changeset/tame-donkeys-hunt.md
+++ b/.changeset/tame-donkeys-hunt.md
@@ -2,4 +2,4 @@
 "@getflip/swirl-components": patch
 ---
 
-Fix `swirl-form-control`'s `inline` prop rendering an overlapping label when the input has a default placeholder (e.g. `swirl-time-input`, `swirl-date-input`). The label is now shown while the input is empty and hidden once it has a value, instead of being keyed off whether the input happens to declare a `placeholder`.
+Fix `swirl-form-control`'s `inline` prop rendering an overlapping label when the input has a default placeholder (e.g. `swirl-time-input`, `swirl-date-input`). While empty, the input's compact single-row layout is preserved and the label fills it like a placeholder; once it has a value, the label now falls back to its normal small pinned position above the value (as it already does for non-inline inputs) instead of overlapping or disappearing.
--- a/.changeset/tame-donkeys-hunt.md
+++ b/.changeset/tame-donkeys-hunt.md
@@ -2,4 +2,4 @@
 "@getflip/swirl-components": patch
 ---
 
-Fix `swirl-form-control`'s `inline` prop rendering an overlapping floating label instead of positioning the label outside the border. `inline` now renders the same layout as `labelPosition="outside"`, which fixes the broken text alignment on `swirl-time-input` (and any other input with a default placeholder, e.g. `swirl-date-input`) when used inline.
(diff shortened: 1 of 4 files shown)


joseabernardes:
https://icy-water-049ec4003-1787.westeurope.3.azurestaticapps.net/?path=/story/components-swirlformcontrol--swirl-form-control&args=inline:!true

image

This change makes the label always be positioned "outside" when we use the `inline` prop, which is a breaking change for all inputs, and might break a lot of our current UI's.

To avoid the breaking change, I see two options. Either we show the `label` or the `placeholder` when `inline` is true.
What would make more sense from a design pov @marcomFlip?

marcomFlip:
Good catch, thanks @joseabernardes, fixed only in Timeinput. Now checking if able to fix in all input fields as well. I will ping you again when ready and ask to review

marcomFlip:
Good catch, thanks for the screenshots @marcomFlip — pushed a fix.

The label was hiding entirely once an inline input got a value, which meant filled dates/times/numbers had no label at all. Now, once the input has a value, the reduced single-row padding no longer applies, and the label falls back to the same small pinned-above-the-value position that non-inline inputs already use correctly — no more overlap, no more disappearing label.

One trade-off to flag: the box does grow slightly once filled (to fit the small label + value), matching how non-inline inputs already look. Verified on Number/Date/Time Control in the SwirlFormControl story — label shows correctly in both empty and filled states, no overlap.

marcomFlip:
Thanks for the reference screenshot — that clarified it, pushed the fix.

\`inline\` now renders the label above the border, in its own row, exactly the same way \`labelPosition="outside"\` already does — for both empty and filled states, no overlap, no disappearing label. Verified against the reference: label + fully bordered box, matching for Number/Date/Time Control with real values typed in.

Note for the team @joseabernardes: this is the change you originally flagged — the label now always renders above the border for **every** existing \`inline\` usage, not just the broken ones. That's now a confirmed design decision rather than an unintended side effect, but it's worth calling out again here since it does change how currently-shipped inline fields look. Changeset bumped to minor to reflect that.

marcomFlip:
Pushed the fix matching your annotated screenshot exactly (commit 776d0b37, 10:42 UTC).

- Empty: label acts as an inside, compact, centered placeholder (Password Control / Select / Autocomplete / Number / Date / Time Control all verified).
- Filled: label moves above the border, same as before.
- Both states use the same box height, so it doesn't jump when the label moves.

Verified with real typing (not synthetic events) in Storybook — Number Control shows the label inside while empty, then moves above the border once "42" is typed, matching your reference 1:1.

joseabernardes:
> https://icy-water-049ec4003-1787.westeurope.3.azurestaticapps.net/?path=/story/components-swirlformcontrol--swirl-form-control&args=inline:!true
> 
> image
> This change makes the label always be positioned "outside" when we use the `inline` prop, which is a breaking change for all inputs, and might break a lot of our current UI's.
> 
> To avoid the breaking change, I see two options. Either we show the `label` or the `placeholder` when `inline` is true. What would make more sense from a design pov @marcomFlip?

The latest fixes still don't solve this issue. From my pov, the `inline` prop shouldn't change the label position, as we rely on the fact that the input has a consistent height throughout its lifecycle, and with this change, that's no longer true, and it affects all inputs, not only the time input.

marcomFlip:
thank you @joseabernardes . Couple of things:
1. "the inline prop shouldn't change the label position", what's the desired outcome then of this prop? I am lost atm
2. If inline prop change the position we can still try to keep "a consistent height throughout its lifecycle". Label moves from inside to outside, input height stays the same with fixed min height. Example from material https://m3.material.io/components/text-fields/specs#09bbf086-55ac-47d0-ad4c-d4a4f53b0f96

PR #1788, opened 2026-09-02
feat: Add translucent variant to swirl-dialog

danizep:
## Summary

Swirl's design direction has been moving overlay surfaces to a "translucent" treatment: a
semi-transparent tinted fill plus a backdrop-filter blur and a light outline, built on the
--s-translucent-* / --s-blur-* token families added in 68007d00, dec171a6 and 8170faef.

swirl-popover gained this in 7b07609a (Add translucent popover variant) and now defaults to
translucent; swirl-card has an opt-in translucent boolean; swirl-menu hardcodes the same
recipe. swirl-dialog was never migrated — .dialog__body is still a fully opaque
--s-surface-overlay-default surface with --s-shadow-level-3.

The goal is to give swirl-dialog the same translucent option, with an API that matches the
existing precedent, so product teams can opt individual dialogs into the newer look without a
visual change for anyone who does not.

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### DIFF
--- a/packages/swirl-components/src/components/swirl-dialog/swirl-dialog.css
+++ b/packages/swirl-components/src/components/swirl-dialog/swirl-dialog.css
@@ -70,15 +70,10 @@
 }
 
 .dialog__body {
+  position: relative;
   z-index: var(--s-z-40);
-  overflow-x: hidden;
-  overflow-y: auto;
+  overflow: hidden;
   width: 18.5rem;
-  max-height: 90vh;
-  padding-top: var(--s-space-24);
-  padding-right: var(--s-space-16);
-  padding-bottom: var(--s-space-16);
-  padding-left: var(--s-space-16);
   border-radius: var(--s-border-radius-l);
   background-color: var(--s-surface-overlay-default);
   text-align: center;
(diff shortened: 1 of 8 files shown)


packages/swirl-components/src/components/swirl-dialog/swirl-dialog.css
```diff
@@ -98,6 +100,13 @@
   }
 }
 
+.dialog--translucent .dialog__body {
+  outline: var(--s-border-width-default) solid
+    var(--s-border-translucent-outline);
+  background-color: var(--s-translucent-low-default);
```
joseabernardes: **thought:** To avoid staking context issues, we usually set the "translucent" background styles in a `::before`. I guess we'll not have those issue here (I don't even remember what they were anymore), but maybe it's worth checking if setting them in a `::before` would also work well here 🤔

joseabernardes (approved):

packages/swirl-components/src/components/swirl-dialog/swirl-dialog.css
danizep: Normally it's when opening popovers, the detection of the boundaries might not work properly.

joseabernardes: Looks good!

PR #1790, opened 2026-09-03
Add focus trigger to swirl popover trigger

danizep:
## Summary

- [Ticket](#)
- [Design](#)
- [Storybook](#)

### DIFF
--- a/packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
+++ b/packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
@@ -1,6 +1,14 @@
 import { Component, Element, h, Host, Prop, Watch } from "@stencil/core";
 import { SwirlPopover } from "../swirl-popover/swirl-popover";
 
+export type SwirlPopoverTriggerMethod = "click" | "hover" | "focus";
+
+const swirlPopoverTriggerMethods: SwirlPopoverTriggerMethod[] = [
+  "click",
+  "hover",
+  "focus",
+];
+
 /**
  * @slot slot - The trigger element (e.g. a button) that opens the popover.
  */
@@ -17,7 +25,10 @@ export class SwirlPopoverTrigger {
   @Prop() parentScrollContainer?: HTMLElement;
   @Prop() setAriaAttributes?: boolean = true;
   @Prop() swirlPopover!: string | HTMLSwirlPopoverElement | SwirlPopover;
+  @Prop() trigger?: SwirlPopoverTriggerMethod[] | string;
+  /** @deprecated Please use the "trigger" prop instead, e.g. trigger="click". */
   @Prop() triggerOnClick?: boolean = true;
+  /** @deprecated Please use the "trigger" prop instead, e.g. trigger="hover focus". */
   @Prop() triggerOnHover?: boolean = false;
   @Prop() hoverLingerDuration?: number;
   @Prop() hoverDelay?: number;
@@ -28,12 +39,18 @@ export class SwirlPopoverTrigger {
   private hoverDelayReference?: NodeJS.Timeout;
   private popoverElRef?: HTMLSwirlPopoverElement;
   private triggerIsActive: boolean = false;
+  private triggerMethods: SwirlPopoverTriggerMethod[] = [];
 
   connectedCallback() {
     if (this.componentLoaded) {
       this.setupHoverListeners();
       this.setupIntersectionObserver();
     }
+  }
+
+  componentWillLoad() {
+    this.warnAboutDeprecatedProps();
+    this.resolveTriggerMethods();
   }
 
   componentDidLoad() {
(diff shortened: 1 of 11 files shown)


packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
```diff
@@ -206,30 +279,122 @@ export class SwirlPopoverTrigger {
...
+
+    return (
+      this.el.contains(node) ||
+      popoverEl === (node as unknown as HTMLElement) ||
+      Boolean(popoverEl?.contains(node))
     );
+  }
+
+  private getActiveElement(): Node | null {
+    return document.activeElement;
```
joseabernardes, line 344: **issue:** We need to use the `getActiveElement` util to ensure we get the active element inside a shadow-root element.

packages/swirl-components/src/components/swirl-popover-trigger/swirl-popover-trigger.tsx
```diff
@@ -206,30 +279,122 @@ export class SwirlPopoverTrigger {
...
+    // open the popover and the click that follows would immediately toggle
+    // it shut again, making the click appear to do nothing.
+    if (
+      this.triggerMethods.includes("click") &&
+      !this.triggerMethods.includes("hover")
+    ) {
+      return;
+    }
+
+    this.openPopover("focus", true);
```
joseabernardes, line 308: **issue:** When we focus on a trigger, the popover never gets focus, which makes it impossible to interact with the popover itself.

danizep: As we discussed before, this change is intended because moving the focus to the popover would break the tab sequence.

> ### AI context:
> No, the trigger element must not lose focus when a popover opens on focus. Automatically shifting keyboard focus away from an element the moment it receives focus disrupts standard keyboard navigation and violates core accessibility standards.
> 
> Why Focus Must Remain on the Trigger
> 
> WCAG 3.2.1 (On Focus - Level A): Receiving focus must never trigger an automatic change of context. Shifting keyboard focus to a new container without explicit user activation (such as pressing Enter or Space) directly violates this rule.
> 
> Predictable Tab Sequence: If a user is pressing Tab to scan through a list of buttons, stealing focus forces their cursor inside the popover. They would have to tab through all popover content before reaching the next button, destroying linear navigation.
> 
> Focus Loops: Automatically moving focus on focus frequently locks keyboard users in an inescapable focus loop or prevents them from reaching the rest of the page.

@Sqrrl you are super versed on this, maybe you have s strong opinion, from what I read ideally we should keep these hover and focus interaction for informative tooltip (not interative) and have the screen reader announce the popover content based on the `aria-describedby`. What do you think?

Or should we ignore screen reader for now and focus mostly on the keyboard navigation.

Sqrrl: That could work for the tooltip use case, but what about interactive popovers? I'm not sure what the best/correct behavior would be.

danizep: I would discourage interactive popovers that open on `focus` or `houver`.

Sqrrl: That sounds fair. Maybe a dumb question, but why can't we use the swirl-tooltip? It already has the focus and aria handling.

joseabernardes: We needed to display a more rich tooltip, with custom content:

image

danizep: It's not a dumb question, but it's because the reason for this change is based on this usage of the popover.

Screenshot 2026-09-15 at 14 35 23

Sqrrl: I see. So either popover with focus support, or tooltip with richtext support. I don't mind using the popover as long as the content is announced when focused.

danizep: To be fair, I think I will leave it as is with the improved keyboard navigation accessibility.
Upon testing I see that the SwirlTooltip also is not announcing it's content, and due to the separation of popover and trigger and the complexities that come with shadow root, the solution it's becoming too complex and potentially not performant with the introduction of MutationObservers.

I think this is a bigger topic worth a deep dive to properly implement the ARIA APG patterns.

joseabernardes: I agree.

The uses of the popover in these conditions can always ensure that its content is announced properly.
That's what is being done in the case Daniel described, where the button that triggers the tooltip already announces the channels that will appear inside the popover.

joseabernardes (approved):

PR #1794, opened 2026-09-07
Interactive data cell

danizep:
## Summary

Adds an interactive, keyboard-navigable variant of swirl-data-cell to support the Figma "Channel cell" design (compact rows with avatar + name + optional tag used in the Post Success Modal), and fixes its rounded-corner behavior when grouped in a swirl-data-cell-stack.

Changes:
- New interactive prop — renders the cell as a native <button> (clickable + keyboard-operable via native Tab/Enter/Space, no custom key handling needed)
- New href / linkTarget props — renders the cell as an <a> when a link target is needed
- New size prop ("s" | "m") — compact (s) sizing matches the Figma channel-cell spec (8px padding, 4px radius, 4px gap)
- Hover/pressed/focus-visible states added for interactive cells across all intents
- Fixed compact (size="s") cells to properly honor first/last-cell corner rounding when grouped inside swirl-data-cell-stack, matching Figma's grouped-list corner treatment (12px outer / 4px touching corners) instead of always rendering flat corners
- 
- [Ticket](#)
- [Design](#)
- [Storybook](#)

### DIFF
--- a/packages/swirl-components/src/components/swirl-data-cell/swirl-data-cell.tsx
+++ b/packages/swirl-components/src/components/swirl-data-cell/swirl-data-cell.tsx
@@ -19,6 +19,8 @@ export type SwirlDataCellIntent =
   | "translucent"
   | "subdued";
 
+export type SwirlDataCellSize = "s" | "m";
+
 /**
  * @slot media - Optional media content (e.g., swirl-avatar, icons). Only swirl-avatar and icon elements are styled.
  * @slot content - Optional content element (e.g., swirl-text-input, swirl-select). When provided, the value prop is ignored. Content automatically takes 100% width when no label is present.
@@ -32,8 +34,12 @@ export type SwirlDataCellIntent =
 export class SwirlDataCell {
   @Element() el: HTMLElement;
 
+  @Prop() href?: string;
   @Prop() intent?: SwirlDataCellIntent = "default";
+  @Prop() interactive?: boolean = false;
   @Prop() label?: string;
+  @Prop() linkTarget?: string;
+  @Prop() size?: SwirlDataCellSize = "m";
   @Prop() tooltip?: string;
   @Prop() value?: string;
   @Prop() vertical?: boolean = false;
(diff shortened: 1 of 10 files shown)


packages/swirl-components/src/components/swirl-data-cell/swirl-data-cell.tsx
```diff
@@ -32,8 +34,12 @@ export type SwirlDataCellIntent =
 export class SwirlDataCell {
   @Element() el: HTMLElement;
 
+  @Prop() href?: string;
   @Prop() intent?: SwirlDataCellIntent = "default";
+  @Prop() interactive?: boolean = false;
   @Prop() label?: string;
+  @Prop() linkTarget?: string;
+  @Prop() size?: SwirlDataCellSize = "m";
```
fabio-carvalho88: **suggestion:** in `swirl-avatar` we have this pattern and I think we should replicate it here

```suggestion
  @Prop({ reflect: true }) size?: SwirlDataCellSize = "m";
```

fabio-carvalho88 (changes requested):

fabio-carvalho88:
**nitpick:** hover styles are propagating to swirl-avatar in the media slot, changing its background. is this the intended behaviour? imo it looks a bit odd

https://github.com/user-attachments/assets/914385f7-bba3-442d-aa55-6c70bb9b4caf

packages/swirl-components/src/components/swirl-data-cell/swirl-data-cell.tsx
```diff
@@ -80,14 +86,15 @@ export class SwirlDataCell {
...
 
-    const className = classnames("data-cell", {
+    const className = classnames("data-cell", `data-cell--size-${this.size}`, {
       "data-cell--vertical": this.vertical,
       "data-cell--has-media": hasMedia,
       "data-cell--has-suffix": hasSuffix,
       "data-cell--has-content": hasContent,
       "data-cell--no-label": !hasLabel,
-      "data-cell--interactive": hasCheckbox || hasRadio,
+      "data-cell--interactive": isInteractive || hasCheckbox || hasRadio,
```
fabio-carvalho88, line 97: **question:** this plus the styles applied mean that all existing checkbox/radio cells change appearance as well. is that intended?

packages/swirl-components/src/components/swirl-data-cell/swirl-data-cell.tsx
danizep: Good point :)

danizep:
> **nitpick:** hover styles are propagating to swirl-avatar in the media slot, changing its background. is this the intended behaviour? imo it looks a bit odd
> 
>  Screen.Recording.2026-09-07.at.15.09.10.mov

Thank you for noticing it. Claude came up with it, i noticed and intended not to commit that part, but  ended up forgetting.

packages/swirl-components/src/components/swirl-data-cell/swirl-data-cell.tsx
danizep: In my opinion this is intentional but I can double check with the design. The previous implementation didn't have hover and pressed for checkbox and radio buttons despite clicking on the whole cell toggling them. At least I see it as an improvement in feedback.

fabio-carvalho88: that's fair! feel free to merge it if design doesn't raise any concerns 🫡

fabio-carvalho88 (approved):

PR #1809, opened 2026-09-18
Add swirlAriaDisabled and swirlAriaDescribedby to swirl-action-list-item

danizep:
swirl-action-list-item's disabled prop renders a native <button disabled>, which is excluded from hit-testing and the accessibility tree: it can't be focused, or tabbed to, so a swirl-tooltip wrapped around a disabled item never opens and swirl-action-list's arrow-key navigation gets stuck on it.

Add a separate, additive swirlAriaDisabled input that renders aria-disabled instead, keeping the item focusable and hoverable while the component itself blocks activation (click, and the synthetic click Enter/Space fire on a focused button). Add a swirlAriaDescribedby input (matching swirl-button's swirlAriaExpanded/swirlAriaHaspopup/swirlAriaDescribedby convention) so a consumer can link the item to a description of why it's disabled, e.g. a tooltip's content.

The existing disabled prop is untouched: every rule keyed on it stays exactly as it was, just extended to also match the new state. No skip logic was added to swirl-action-list's keyboard navigation — swirl-option-list already establishes that disabled items should stay focusable and only block activation, which also happens to fix the pre-existing "arrow keys get stuck on a disabled item" bug for the new state.

## Summary

- [Ticket](#)
- [Design](#)
- [Storybook](#)


packages/swirl-components/src/components/swirl-action-list-item/swirl-action-list-item.tsx
```diff
@@ -26,6 +26,8 @@ export class SwirlActionListItem {
   @Prop() intent?: SwirlActionListItemIntent = "default";
   @Prop() label!: string;
   @Prop() size?: SwirlActionListItemSize = "m";
+  @Prop() swirlAriaDescription?: string;
```
danizep: @Sqrrl initially I went for `aria-describedby` but because the referenced element is not in the shadow-root it cannot be referenced. This made the `aria-description` the more obvious solution or alternatively programmatically set `ariaDescribedByElements`.

Sqrrl: Could you research the real-world support of `aria-description` a bit? The browser support is fine, but I'm not sure assistive technology has good enough support yet.

danizep: Apparently already works on NVDA, JAWS, VoiceOver (iOS/macOS), Narrator, and TalkBack, but since it's still in draft I changed back to  aria-describedby programmatically set.

packages/swirl-components/src/components/swirl-action-list-item/swirl-action-list-item.tsx
```diff
@@ -42,10 +55,37 @@ export class SwirlActionListItem {
...
+    }
+
+    const root = this.el.getRootNode() as Document | ShadowRoot;
+    const elements = (this.swirlAriaDescribedby ?? "")
+      .split(/\s+/)
+      .filter(Boolean)
+      .map((id) => root.getElementById?.(id))
+      .filter((element): element is HTMLElement => Boolean(element));
+
+    this.buttonEl.ariaDescribedByElements =
```
Sqrrl, line 76: **suggestion:** Could we check if the value is actually different to avoid unnecessary DOM mutations on every render?

Sqrrl (approved):

packages/swirl-components/src/components/swirl-action-list-item/swirl-action-list-item.tsx
Sqrrl: Probably a good idea to go with the conservative approach here.

packages/swirl-components/src/components/swirl-action-list-item/swirl-action-list-item.tsx
danizep: Good point. I opted for adding a `@Watch("swirlAriaDescribedby")` instea of having a check that runs on `componentDidRender`.

